PR #8, opened 2026-05-10
Fix/GitHub oauth signin

### LINKED ISSUE #7: [BUG] GitHub sign-in route returns 404
## Describe the Bug

Clicking "Sign in with GitHub" on the homepage routes to `/api/auth/signin`, but the app does not currently define the NextAuth API route needed to handle that path. This causes the sign-in page to return a 404.

The project includes `next-auth` and documents GitHub OAuth support, so the sign-in button appears to be pointing at an intended but missing auth route.

## Steps to Reproduce

1. Start the client app.
2. Go to `http://localhost:3000`.
3. Click on "Sign in with GitHub".
4. See the 404 page at `http://localhost:3000/api/auth/signin`.

## Expected Behavior

Clicking "Sign in with GitHub" should start the GitHub OAuth flow and allow the user to authenticate with GitHub.

## Screenshots / Logs

Browser URL:

```text
http://localhost:3000/api/auth/signin
```

Error shown:

```text
404 This page could not be found.
```

## Environment

- OS: Windows
- Node version: v22.13.1
- Browser: Chrome

## Additional Context

The README mentions GitHub OAuth via NextAuth, and `next-auth` is listed in the client dependencies. However, the App Router route for NextAuth does not appear to be present in the current codebase.

Chris8115:
## Summary

Fixes GitHub sign-in by adding the missing NextAuth API route for the App Router and sending users to the dashboard after authentication.

Closes #7 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / code cleanup

## Changes Made

- Added the missing NextAuth route at `client/src/app/api/auth/[...nextauth]/route.ts`.
- Added GitHub provider configuration in `client/src/lib/auth.ts`.
- Updated the homepage sign-in link to use the GitHub provider directly and redirect to `/dashboard` after login.

## How to Test

Steps for the reviewer to verify this works:

1. Set `NEXTAUTH_URL`, `NEXTAUTH_SECRET`, `GITHUB_ID`, and `GITHUB_SECRET` in `client/.env.local`.
2. Set the GitHub OAuth callback URL to `http://localhost:3000/api/auth/callback/github`.
3. Start the client app with `npm run dev`.
4. Open `http://localhost:3000`.
5. Click “Sign in with GitHub”.
6. Complete GitHub authorization.
7. Confirm the app redirects to `/dashboard`.
8. Visit `http://localhost:3000/api/auth/session` and confirm a session object is returned.

## Screenshots (if UI change)

N/A

## Checklist

- [x] Linked issue in summary
- [x] `npm run lint` passes locally
- [x] Self-reviewed the diff
- [x] Added/updated tests if applicable


Priyanshu-byte-coder (commented):
Thanks for the contribution! The core fix (adding the missing NextAuth route) is correct and needed. But there are a few issues to resolve before this can merge.

## Blockers

### 1. Two competing auth systems

This PR adds NextAuth (`client/src/lib/auth.ts`) **and** keeps the Express backend OAuth flow (`server/src/controllers/githubController.ts`). They both handle GitHub OAuth but with different callback URLs and token formats — they'll conflict.

The sign-in button now uses NextAuth's flow (`/api/auth/signin/github?callbackUrl=/dashboard`), which calls back to `/api/auth/callback/github` (NextAuth's route). But `AuthTokenHandler` looks for a `#token=` hash, which only comes from the Express backend's redirect in `githubCallback`. With the current setup, the hash will never appear and the component does nothing.

**Please pick one flow and remove the other:**
- **Option A (recommended):** Use NextAuth fully — remove `githubSignIn`, remove `AuthTokenHandler`, add a `callbacks.session` in `auth.ts` to expose the GitHub access token via `useSession()`
- **Option B:** Keep Express JWT flow — remove `auth.ts`, remove the NextAuth route, revert the sign-in link to point to the Express backend

### 2. JWT token passed via URL hash is a security risk

`res.redirect(`${clientUrl}/dashboard#token=${...}`)` puts a JWT in the URL hash. While hashes don't go to the server, they:
- Are stored in browser history (visible to anyone with access to the machine)
- Are readable by browser extensions
- May appear in analytics tools

If keeping the Express JWT flow (Option B), exchange the token via a short-lived `httpOnly` cookie or a one-time code instead.

### 3. Unused variable

In `client/src/app/page.tsx`:
```ts
const apiUrl = process.env.NEXT_PUBLIC_API_URL ?? "http://localhost:4000";
```
`apiUrl` is declared but never used. Remove it.

## Minor

- `auth.ts` is missing a `callbacks` block — without it `useSession()` won't expose the GitHub access token, so components can't call the GitHub API on behalf of the user. If going with Option A, add:

```ts
callbacks: {
  async session({ session, token }) {
    session.accessToken = token.accessToken as string;
    return session;
  },
  async jwt({ token, account }) {
    if (account) token.accessToken = account.access_token;
    return token;
  },
},
```

Fix the two blockers and this is close to mergeable. Let me know which auth option you want to go with and I can give more specific guidance.

Chris8115:
Thanks for the detailed review. I went with Option A and moved the app to the NextAuth-only flow.

Changes made:

- Removed the Express GitHub OAuth route/controller so there is no competing callback flow.
- Removed AuthTokenHandler and the #token= handling path.
- Added NextAuth jwt and session callbacks so the GitHub access token is exposed as session.accessToken.
- Added SessionProvider and NextAuth session type augmentation for accessToken.
- Removed the unused apiUrl variable from client/src/app/page.tsx.
- Cleaned the README and env examples so they no longer reference the old backend OAuth/JWT flow.

I also ran type checks for both client and server, and both pass. next lint is currently blocked by the repo not having ESLint initialized yet, since it opens the interactive setup prompt.

PR #111, opened 2026-05-15
fix: show inline error when goal creation fails

Harikishanth:
## Summary

Show an inline error message below the submit button when goal creation
fails, instead of silently resetting the form.

Closes #90

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / code cleanup

## Changes Made

- Added `createError` state (`string | null`) to `GoalTracker.tsx`
- Clear `createError` at the start of each submission attempt
- Scoped `try/catch` to only the POST request so a `loadGoals()` failure after a successful creation doesn't show a false error
- Render error message below the submit button using `text-red-500`

## How to Test

Steps for the reviewer to verify this works:

1. Sign in and go to the dashboard
2. Open DevTools → Network tab → set to Offline
3. Type a goal label, set a target, click "Add goal"
4. Confirm "Failed to create goal. Please try again." appears below the button
5. Re-enable network, submit again — confirm error clears and goal is created

## Screenshots (if UI change)

N/A — error text appears below the submit button on failure

## Checklist

- [x] Linked issue in summary
- [x] `npm run lint` passes locally
- [x] No TypeScript errors (`npm run type-check`)
- [x] Self-reviewed the diff
- [ ] Added/updated tests if applicable

PR #106, opened 2026-05-15
feat(goals): add delete button to individual goals in GoalTracker (#98)

26Naitik:
## Summary
Closes #98

Adds a delete button to individual goals in the `GoalTracker` component, allowing users to remove goals they no longer want to track.

*(Note: This is a clean re-submission of the previous PR, containing only the 2 files relevant to this feature without any merge conflicts or unrelated files).*

## Changes
- **New**: `src/app/api/goals/[id]/route.ts` — `DELETE` endpoint with session authentication and a user ownership guard (ensuring users can only delete their own goals).
- **Modified**: `src/components/GoalTracker.tsx` — Added a trash icon button, inline Yes/No confirmation state, and optimistic UI updates for instantaneous feedback.

## Behavior
1. A 🗑 trash icon appears next to each goal.
2. Clicking the trash icon shows an inline **"Delete? Yes / No"** prompt (avoiding disruptive `window.confirm` dialogs).
3. Clicking **Yes** removes the goal optimistically from the UI and calls `DELETE /api/goals/{id}`.
4. If the API call fails, the goal is automatically restored to the list.
5. Buttons are disabled while the delete request is in-flight to prevent duplicate requests.

## Checklist
- [x] Tested the optimistic UI flow and error recovery.
- [x] Follows existing authentication and Supabase patterns from `goals/route.ts`.
- [x] No `desktop.ini` or unrelated files are included in this PR.

PR #114, opened 2026-05-15
added export feature as pdf and csv

### LINKED ISSUE #16: [FEAT] Export metrics to CSV/PDF
## What Needs to Be Done

Add an export button to the dashboard that downloads metrics (contributions, PRs, goals) as a CSV or PDF file.

## Why This Matters

Developers may want to share their metrics in standups, include in portfolios, or import into spreadsheets for further analysis.

## Files to Create/Modify

- `src/components/ExportButton.tsx` — new client component with dropdown (CSV / PDF)
- `src/app/dashboard/page.tsx` — add ExportButton to header
- CSV generation: can be done client-side from the already-fetched data
- PDF: use a library like \`jspdf\` or \`@react-pdf/renderer\`

## Acceptance Criteria

- [ ] Export button visible on dashboard
- [ ] CSV export works for contributions (date, count) and PR stats
- [ ] File downloads with a readable filename like \`devtrack-metrics-2025-05-10.csv\`
- [ ] No TypeScript errors, lint passes

## Tech Context

For CSV (simpler, start here):
```ts
const csv = "Date,Commits\n" + data.map(d => \`\${d.day},\${d.commits}\`).join("\n");
const blob = new Blob([csv], { type: "text/csv" });
const url = URL.createObjectURL(blob);
// trigger download via <a> element
```

PDF can be a follow-up PR.

PaperCodeGithub:
## Summary

Brief description of what this PR does.

Closes #16 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [x] Refactor / code cleanup

## Changes Made
- Add export button to metrics dashboard 
- Moved the `fetch()` call out of `PRMetrics.tsx`, `GoalTracker.tsx` and `ContributionGraph.tsx` up into `page.tsx` then pass the props to export function.
- Added new libraries `jspdf` for generating PDF documents and `jspdf-autotable` for formating data into clean, good-looking tables

## How to Test

Steps for the reviewer to verify this works:

1. Open the dashboard page
2. Click on Export CSV or Export PDF button

## Screenshots (if UI change)
image
Buttons are attachs as shown in Screen Shot

## Checklist

- [x] Linked issue in summary
- [x] `npm run lint` passes locally
- [x] Self-reviewed the diff
- [ ] Added/updated tests if applicable


Priyanshu-byte-coder (changes requested):
Really good work on the architecture — lifting the fetches into `page.tsx` and passing data down as props is exactly the right approach, and the CSV generation is clean. A few things to fix before this can merge:

## Blockers

### 1. `NEXT_PUBLIC_APP_URL` falls back to `localhost:3000` on Vercel
```js
const baseUrl = process.env.NEXT_PUBLIC_APP_URL || "http://localhost:3000";
```
`NEXT_PUBLIC_APP_URL` is not set in the Vercel environment, so the server-side fetches in `page.tsx` will hit `localhost:3000` and fail silently — `prData` and `goalsData` will always be null/empty in production. Use `NEXTAUTH_URL` as the fallback instead (it's already set):
```js
const baseUrl = process.env.NEXT_PUBLIC_APP_URL || process.env.NEXTAUTH_URL || "http://localhost:3000";
```

### 2. Hardcoded dark colors break light mode
`GoalTracker.tsx`, `PRMetrics.tsx`, and `ExportButton.tsx` use hardcoded Tailwind colors that look broken in light mode. Replace with the CSS variables already defined in `globals.css`:

| Hardcoded | Should be |
|---|---|
| `bg-slate-800` | `bg-[var(--card)]` |
| `bg-slate-700` | `bg-[var(--card-muted)]` |
| `bg-indigo-600` / `hover:bg-indigo-500` | `bg-[var(--accent)]` / `hover:opacity-90` |
| `text-white` | `text-[var(--card-foreground)]` |
| `text-slate-400` | `text-[var(--muted-foreground)]` |

### 3. PRMetrics lost its error state
The original component showed an error message when the fetch failed. The refactored version uses `{!metrics ? <skeleton> : ...}` — if the API call fails, `metrics` stays null and the loading skeleton shows indefinitely. Add a visible error state for when `metrics` is null after load.

### 4. Merge conflicts
The PR is currently conflicting with main — please rebase onto the latest main and resolve conflicts.

---

The core implementation is solid and the pagination note about `per_page=100` is a good catch — do open that separate issue as discussed!

Priyanshu-byte-coder (changes requested):
Three things still blocking:

## 1. Still conflicting
The PR is marked as conflicting with main — the conflict resolution commit didn't fully resolve it. Rebase onto current main again.

## 2. `package-lock.json` still incorrectly modified
The diff still removes `"peer": true` from existing packages (12 deletions) while adding jspdf entries. The `"peer": true` lines must stay untouched. Run:
```bash
git checkout origin/main -- package-lock.json
npm install
git add package-lock.json
git commit -m "fix: regenerate package-lock from clean base"
```
This will produce a lockfile that only adds the jspdf entries without touching existing ones.

## 3. GoalTracker and PRMetrics prop-drilling now conflicts with merged PRs
Since the last review, two PRs have been merged that modify these files:
- **#111** — added inline error state to GoalTracker's create form
- **#106** — added delete button with inline confirm to GoalTracker

The approach of making GoalTracker prop-driven (removing fetch + loading state) will now conflict with both of these changes. Consider a simpler path: keep GoalTracker and PRMetrics fetching their own data (as they do now), and have `ExportButton` make its own fetch calls for the export data. This avoids the architectural change entirely and the PR becomes much smaller and focused.

## 4. `NEXT_PUBLIC_APP_URL` fallback still missing
The server-side fetches in `page.tsx` still use:
```js
const baseUrl = process.env.NEXT_PUBLIC_APP_URL || "http://localhost:3000";
```
Use `process.env.NEXTAUTH_URL` as the middle fallback — it's already set in Vercel.

PaperCodeGithub:
As asked
- Resolved conflicts
- Added `process.env.NEXTAUTH_URL` as the middle fallback
- Export feature
- moved `fetch()` calls from components to `dashboard/page.tsx`

PaperCodeGithub:
My local npm version keeps aggressively stripping out the peer: true lines whenever I run npm install. To keep the PR clean and avoid lockfile noise, I've reverted package-lock.json to match main exactly. Could you generate the lockfile update on your end when merging?

Priyanshu-byte-coder (changes requested):
Good progress — package-lock is clean, `NEXTAUTH_URL` fallback is in place, conflict resolved. Two small fixes needed before merge:

**1. Hardcoded button colors in ExportButton**

```tsx
className="px-4 py-2 bg-slate-700 hover:bg-slate-600 text-white ..."
className="px-4 py-2 bg-indigo-600 hover:bg-indigo-500 text-white ..."
```

These don't respect the app theme. Please use CSS vars:
- CSV button: `bg-[var(--control)] border border-[var(--border)] text-[var(--card-foreground)] hover:border-[var(--accent)]`
- PDF button: `bg-[var(--accent)] text-[var(--accent-foreground)] hover:opacity-90`

**2. Missing `type="button"` on both buttons in ExportButton**

Without this, if the component ever ends up inside a `<form>`, clicking will submit the form. Add `type="button"` to both.

Everything else looks good — the server-side data fetch, the NEXTAUTH_URL fallback, the export logic itself.

PaperCodeGithub:
As requested, I added `type="button'` and used mentioned styling

Priyanshu-byte-coder:
The CSS variable theming and `type="button"` fixes from the last review are in — good work. One layout regression still blocking merge:

**Row 2 grid layout removed**

The diff changes `<div className="mt-6 grid grid-cols-1 lg:grid-cols-2 gap-6">` to `<div className="mt-6">`. This means `PRBreakdownChart` no longer sits side-by-side with `PRMetrics` — both stack vertically instead. The 2-column grid is intentional in the current layout.

Fix: keep the grid classes on Row 2's wrapper div:
```tsx
{/* Row 2: PR metrics + PR breakdown */}
<div className="mt-6 grid grid-cols-1 lg:grid-cols-2 gap-6">
  <PRMetrics metrics={prData} />
  <PRBreakdownChart />
</div>
```

Once that's fixed, this should be ready to merge.

Priyanshu-byte-coder (changes requested):
Good progress — CSS vars and `type="button"` are fixed. One architectural issue remains:

**Server-side data fetching in `dashboard/page.tsx`**

```ts
const fetchOptions = { headers: headers(), cache: "no-store" };
const [prRes, goalsRes, contribRes] = await Promise.all([
  fetch(`${baseUrl}/api/metrics/prs`, fetchOptions),
  ...
]);
```

Every other widget on the dashboard fetches its own data client-side. This approach fetches data in the server component and passes it as props to `ExportButton` — which breaks that pattern and makes the dashboard heavier for all users even when they never export.

**Please make `ExportButton` self-fetching:**

```tsx
// Inside ExportButton.tsx
const [prData, setPrData] = useState(null);
const [contribData, setContribData] = useState([]);

useEffect(() => {
  fetch('/api/metrics/prs').then(r => r.json()).then(setPrData);
  fetch('/api/metrics/contributions?days=365').then(r => r.json()).then(setContribData);
}, []);
```

Fetch only when the user clicks export (lazy), or on mount. Either way, keep it inside the client component. Remove the server-side fetches and `import { headers } from "next/headers"` from `dashboard/page.tsx`.

The `jsPDF` usage for PDF generation is fine.

PaperCodeGithub:
Hey! No worries at all. We actually went with the server-side prop-drilling approach because we agreed on 'Option 3' back in the original issue thread to avoid duplicate client requests.

That said, I will revert `page.tsx` back to its original clean state. Instead of fetching on mount with `useEffect` (which would still slow down the page), I'll make `ExportButton` do a lazy fetch, it will only hit the APIs concurrently inside the onClick handler when the user actually clicks 'Export'.

Priyanshu-byte-coder (approved):
Addressed. Fetch is now client-side inside ExportButton, no props passed from server, dashboard/page.tsx stays a thin auth-only server component. Merging.

PR #328, opened 2026-05-18
fix: prevent Share Profile button flash before settings load

### LINKED ISSUE #194: [BUG] DashboardHeader 'Share Profile' link shown before is_public setting loads
## Description
The **Share Profile** button in DashboardHeader flashes visible briefly during the initial `useEffect` fetch, even for users who have `is_public = false`. This is because the initial state is `null` or `false` but the condition check allows a brief render.

## Steps to Reproduce
1. Sign in with a user whose public profile is disabled
2. Hard refresh the dashboard
3. Observe the Share Profile button flashing visible for ~100–200 ms before disappearing

## Root Cause
```tsx
const [isPublic, setIsPublic] = useState(false); // renders as hidden
// BUT during fetch, if conditional is wrong:
{isPublic && <a>Share Profile</a>}
```

The flash may occur if `isPublic` starts as `null` or `undefined` before the fetch resolves.

## Fix
Use a tri-state: `null` = loading, `true/false` = loaded. Only render the button when state is explicitly `true`:
```tsx
const [isPublic, setIsPublic] = useState<boolean | null>(null);
{isPublic === true && <a>Share Profile</a>}
```

## Files to Touch
`src/components/DashboardHeader.tsx`

jaygaikar-09:
## Summary

Fixes the issue where the **"Share Profile"** button briefly flashed on dashboard load for users with private profiles because the UI rendered before `/api/user/settings` finished loading.

Closes #194

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / code cleanup

---

## Changes Made

- Replaced the settings-based visibility check with a tri-state `isPublic` flag:
  - `null` → loading
  - `true` → public profile
  - `false` → private profile or fetch failure
- Updated rendering logic to only show the button when:
  ```tsx
  isPublic === true && session?.githubLogin

### DIFF
--- a/src/components/DashboardHeader.tsx
+++ b/src/components/DashboardHeader.tsx
@@ -7,26 +7,28 @@ import SignOutButton from "@/components/SignOutButton";
 import ThemeToggle from "@/components/ThemeToggle";
 import UserAvatar from "@/components/UserAvatar";
 
-interface UserSettings {
-  is_public: boolean;
-}
-
 export default function DashboardHeader() {
   const { data: session } = useSession();
-  const [settings, setSettings] = useState<UserSettings | null>(null);
+  const [isPublic, setIsPublic] = useState<boolean | null>(null);
 
   useEffect(() => {
-    if (!session) return;
+    if (!session) {
+      setIsPublic(null);
+      return;
+    }
 
     async function loadSettings() {
       try {
         const res = await fetch("/api/user/settings");
         if (res.ok) {
           const data = await res.json();
-          setSettings(data);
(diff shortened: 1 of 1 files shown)


Priyanshu-byte-coder (approved):
Clean fix. Changing to `isPublic === true` strict check ensures the button stays hidden until settings are confirmed loaded, not just until the first API call starts. Error/failure paths now default to `false` instead of leaving `null` indefinitely. The flash is eliminated.

PR #338, opened 2026-05-19
fix(security): prevent rate limit bypass via X-Forwarded-For spoofing

### LINKED ISSUE #337: [BUG] Public profile rate limiter bypassed via X-Forwarded-For header spoofing
## Summary

The rate limiter in \`src/app/api/public/[username]/route.ts\` uses \`x-forwarded-for\` as the primary key for tracking request counts. This header is fully client-controlled — any caller can set it to an arbitrary value on every request, making the 30 req/min limit completely ineffective.

## Steps to Reproduce

\`\`\`bash
for i in $(seq 1 200); do
  curl -H "X-Forwarded-For: $i.0.0.1" \
    https://<deployed-url>/api/public/someuser
done
\`\`\`

All 200 requests succeed. The 429 limit is never triggered.

## Root Cause

\`getRateLimitKey()\` trusts \`x-forwarded-for\` first:

\`\`\`typescript
function getRateLimitKey(req: NextRequest): string {
  return (
    req.headers.get("x-forwarded-for") || // client-controlled, spoofable
    req.headers.get("x-real-ip") ||
    req.ip ||
    "unknown"
  );
}
\`\`\`

## Impact

An unauthenticated attacker can exhaust the shared \`GITHUB_TOKEN\` quota (5,000 req/hr) in minutes by rotating spoofed IPs while iterating public usernames. Once exhausted, all authenticated users receive API errors until the quota resets.

## Fix

Use \`req.ip\` as the primary key. The Next.js/Vercel runtime sets this from the verified network-layer source address and it cannot be set by the caller.

\`\`\`typescript
function getRateLimitKey(req: NextRequest): string {
  return req.ip || req.headers.get("x-real-ip") || "unknown";
}
\`\`\`

## GSSoC Info

Difficulty: Intermediate | Type: Bug / Security

anshul23102:
Closes #337

## What was wrong

`getRateLimitKey()` in `src/app/api/public/[username]/route.ts` used `x-forwarded-for` as the primary source for the rate limit key. This header is a plain HTTP request header — any caller can set it to an arbitrary value on every request. This made the 30 req/min in-memory rate limiter completely ineffective against anyone who rotated the header, allowing unbounded requests to the public profile endpoint and full exhaustion of the shared `GITHUB_TOKEN` quota (5,000 req/hr).

## What changed

- Removed `x-forwarded-for` from the rate limit key lookup chain
- `req.ip` is now the primary key — the Next.js/Vercel runtime populates this from the verified network-layer source address, which cannot be set by the caller
- `x-real-ip` is kept as a fallback for non-Vercel self-hosted deployments
- Added an explanatory comment in the function so the intentional exclusion of `x-forwarded-for` is clear to future contributors

## Files changed

- `src/app/api/public/[username]/route.ts` — `getRateLimitKey()` function only

## Testing

The fix can be verified locally by running the endpoint with a custom `X-Forwarded-For` header after more than 30 requests — the 429 response should now trigger based on the actual connection IP regardless of the header value.

## Type of change

- [x] Bug fix
- [x] Security improvement

### DIFF
--- a/src/app/api/public/[username]/route.ts
+++ b/src/app/api/public/[username]/route.ts
@@ -19,12 +19,15 @@ const RATE_LIMIT_REQUESTS = 30;
 const RATE_LIMIT_WINDOW_MS = 60 * 1000; // 1 minute
 
 function getRateLimitKey(req: NextRequest): string {
-  return (
-    req.headers.get("x-forwarded-for") ||
-    req.headers.get("x-real-ip") ||
-    req.ip ||
-    "unknown"
-  );
+  // req.ip is populated by the Next.js / Vercel runtime from the verified
+  // network-layer source address and cannot be spoofed by the caller.
+  //
+  // x-forwarded-for is intentionally excluded here: it is a plain request
+  // header that any client can set to an arbitrary value. Trusting it as the
+  // primary key allows an attacker to rotate the header on every request,
+  // bypass the per-IP limit entirely, and exhaust the shared GITHUB_TOKEN
+  // quota (5 000 req/hr), making the endpoint unavailable for all users.
+  return req.ip || req.headers.get("x-real-ip") || "unknown";
 }
 
 function checkRateLimit(ip: string): { allowed: boolean; retryAfter?: number } {


Priyanshu-byte-coder (approved):
Correct. `x-forwarded-for` is a plain request header — any caller can set it arbitrarily per request, completely bypassing the rate limit. `req.ip` is populated by the Vercel/Next.js runtime from the verified network-layer source address and cannot be spoofed. Clean one-liner fix.

PR #342, opened 2026-05-19
fix(security): bind link-github OAuth state to authenticated session

### LINKED ISSUE #341: [BUG] link-github OAuth state not bound to session — attacker can link their GitHub account to another user's profile
## Summary

The OAuth \`state\` generated by \`/api/auth/link-github\` is a random nonce with no binding to the authenticated user's session. This means an attacker who can place their state cookie on a victim's browser can trick the callback into running under the victim's session, causing the attacker's secondary GitHub account to be linked to the victim's devtrack profile.

## Root Cause

**Initiation handler** (\`src/app/api/auth/link-github/route.ts\`):
\`\`\`typescript
const state = randomBytes(32).toString("hex"); // no session binding
\`\`\`

**Callback handler** (\`src/app/api/auth/link-github/callback/route.ts\`):
\`\`\`typescript
if (!stateCookie || !state || stateCookie !== state) { /* reject */ }
// Session is checked AFTER state — but state has no user identity in it
const session = await getServerSession(authOptions);
\`\`\`

The state cookie is validated against the URL parameter, but because the state contains no session identity, it provides no protection if the cookie is replayed under a different user's session.

## Attack Scenario

1. Attacker (signed in as user A) visits \`/api/auth/link-github\` — browser gets cookie \`link_github_state=XYZ\` where \`XYZ\` is the attacker's state
2. Attacker places cookie \`link_github_state=XYZ\` on victim's browser (via XSS, cookie tossing on a shared subdomain, or a shared device)
3. Attacker completes GitHub OAuth with their own secondary account, receives \`code=ABC&state=XYZ\` from GitHub
4. Attacker directs victim to \`/api/auth/link-github/callback?code=ABC&state=XYZ\`
5. Callback: cookie matches state (XYZ == XYZ) — passes. Session = victim's session. The attacker's secondary GitHub account is now linked to the victim's devtrack profile.

## Impact

- Attacker's GitHub metrics appear merged with the victim's dashboard
- Attacker can pollute or inflate the victim's contribution data
- Victim loses control over which accounts are linked to their profile

## Fix

Embed \`session.githubId\` into the state (\`<nonce>.<githubId>\`) during initiation. In the callback, after the cookie/state match, extract and verify the embedded ID equals \`session.githubId\`. A state generated for user A is then rejected when the callback runs under user B's session.

## GSSoC Info

Difficulty: Advanced | Type: Security

anshul23102:
Closes #341

## What was wrong

The OAuth \`state\` in \`/api/auth/link-github\` was a random nonce with no binding to the authenticated user's session. An attacker who places their state cookie on a victim's browser can redirect the victim to the callback URL with the attacker's OAuth code. The callback passes the cookie/state match check and links the attacker's secondary GitHub account to the victim's devtrack profile under the victim's session.

## What changed

**\`src/app/api/auth/link-github/route.ts\`**
- State is now \`<nonce>.<session.githubId>\` instead of just a random nonce
- The session user's GitHub ID is embedded so the callback can verify it

**\`src/app/api/auth/link-github/callback/route.ts\`**
- After the existing cookie/state match check, extracts the embedded githubId from the state
- Verifies the embedded ID equals \`session.githubId\`
- Rejects with \`invalid_state\` if they differ — a state generated for user A cannot complete the flow under user B's session

## Why this is safe

GitHub IDs are numeric integers and will never contain a \`.\`, so the \`<nonce>.<githubId>\` format parses unambiguously. The 32-byte random nonce prefix still prevents brute-force state guessing.

## Files changed

- \`src/app/api/auth/link-github/route.ts\`
- \`src/app/api/auth/link-github/callback/route.ts\`

## Type of change

- [x] Security fix
- [x] Bug fix

### DIFF
--- a/src/app/api/auth/link-github/callback/route.ts
+++ b/src/app/api/auth/link-github/callback/route.ts
@@ -44,6 +44,19 @@ export async function GET(req: NextRequest) {
     );
   }
 
+  // Verify the state was generated for this session's user.
+  // The state format is `<nonce>.<githubId>` (set in the initiation handler).
+  // Without this check, an attacker who places their state cookie on a
+  // victim's browser can trick the callback into linking the attacker's
+  // secondary GitHub account to the victim's devtrack profile.
+  const embeddedGithubId = state.split(".").slice(1).join(".");
+  if (!embeddedGithubId || embeddedGithubId !== session.githubId) {
+    return NextResponse.redirect(
+      buildSettingsRedirect("error", "invalid_state"),
+      { status: 302 }
+    );
+  }
+
   const redirectUri = `${process.env.NEXTAUTH_URL ?? ""}/api/auth/link-github/callback`;
 
   const tokenResponse = await fetch("https://github.com/login/oauth/access_token", {
--- a/src/app/api/auth/link-github/route.ts
+++ b/src/app/api/auth/link-github/route.ts
@@ -13,7 +13,14 @@ export async function GET() {
     );
   }
 
-  const state = randomBytes(32).toString("hex");
+  // Bind the random nonce to the authenticated session by appending the
+  // session user's GitHub ID. The callback verifies that the ID embedded
+  // in the state matches the session that completes the flow, so a
+  // state cookie placed on a different user's browser (e.g. via XSS or
+  // cookie tossing) cannot be used to link an attacker's GitHub account
+  // to the victim's devtrack profile.
+  const nonce = randomBytes(32).toString("hex");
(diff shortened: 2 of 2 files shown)


Priyanshu-byte-coder (approved):
Solid CSRF mitigation. Embedding `githubId` in the state nonce and verifying it in the callback prevents cookie-tossing attacks — an attacker's state cookie placed on a victim's browser will fail the ID check. The nonce still provides randomness so state can't be predicted. Correct.

PR #345, opened 2026-05-19
fix(goals): prevent race condition on period reset and fix UTC boundary

### LINKED ISSUE #344: [BUG] Goal period reset has race condition and uses local timezone instead of UTC
## Summary

The \`GET /api/goals\` handler in \`src/app/api/goals/route.ts\` has two bugs that cause goal progress to be incorrectly lost at period boundaries.

## Bug 1 — Race condition on concurrent period resets

\`Promise.all\` fires all goal-reset DB writes concurrently. When two requests arrive simultaneously (two browser tabs or mobile + desktop), both read the same stale \`period_start\`, both decide a reset is needed, and both issue \`UPDATE current=0\`. Any progress incremented between the two reads is silently zeroed by the second write.

\`\`\`typescript
const processedGoals = await Promise.all(
  (goals ?? []).map(async (goal) => {
    // Both concurrent requests see stale period_start
    if (storedPeriodStart < periodStart) {
      await supabaseAdmin.from("goals")
        .update({ current: 0, ... }) // second write clobbers progress
        .eq("id", goal.id)
    }
  })
);
\`\`\`

**Fix:** Add \`.lt("period_start", periodStart)\` to the update so only one concurrent write wins. The losing request re-fetches the current row instead of returning a stale value.

## Bug 2 — Local timezone used for period boundary

\`getPeriodStart\` uses \`getDay()\`, \`setDate()\`, and \`setHours()\` which all operate in the server's local timezone. On servers not running in UTC, the weekly Monday boundary and monthly first-of-month boundary shift by the UTC offset — resetting goals a day early or late.

\`\`\`typescript
const day = now.getDay();       // local timezone
monday.setDate(now.getDate() + diff);   // local timezone
monday.setHours(0, 0, 0, 0);           // local timezone

new Date(now.getFullYear(), now.getMonth(), 1, 0, 0, 0, 0) // local timezone
\`\`\`

**Fix:** Switch to \`getUTCDay()\`, \`setUTCDate()\`, \`setUTCHours()\`, and \`Date.UTC()\` so the boundary is always calculated in UTC.

## GSSoC Info

Difficulty: Advanced | Type: Bug / Data integrity

anshul23102:
Closes #344

## What was wrong

Two bugs in \`src/app/api/goals/route.ts\` caused goal progress to be lost at period boundaries.

**Bug 1 - Race condition:**
\`Promise.all\` fires all goal-reset writes concurrently. Two simultaneous GET requests both read the same stale \`period_start\`, both issue \`UPDATE current=0\`, and the second write zeros out any progress recorded between the two reads.

**Bug 2 - Local timezone boundary:**
\`getPeriodStart\` used local-timezone methods (\`getDay\`, \`setDate\`, \`setHours\`, \`Date(year, month, ...)\`). On servers not running in UTC the weekly and monthly reset boundaries shift by the UTC offset, triggering resets a day early or late.

## What changed

- Added \`.lt("period_start", periodStart)\` to the reset update so only one concurrent write wins. The losing write re-fetches the current DB row instead of returning stale data.
- Switched \`getPeriodStart\` to UTC methods (\`getUTCDay\`, \`setUTCDate\`, \`setUTCHours\`, \`Date.UTC\`) so boundaries are timezone-independent.

## Files changed

- \`src/app/api/goals/route.ts\`

## Type of change

- [x] Bug fix
- [x] Data integrity improvement

### DIFF
--- a/src/app/api/goals/route.ts
+++ b/src/app/api/goals/route.ts
@@ -9,15 +9,21 @@ type Recurrence = "none" | "weekly" | "monthly";
 function getPeriodStart(recurrence: Recurrence): string {
   const now = new Date();
   if (recurrence === "weekly") {
-    const day = now.getDay();
+    // Use UTC methods so the Monday boundary is the same regardless of the
+    // server's local timezone. getDay() / setDate() / setHours() all operate
+    // in local time, which can push the reset boundary a day early or late
+    // on servers that are not running in UTC.
+    const day = now.getUTCDay();
     const diff = day === 0 ? -6 : 1 - day; // Monday
     const monday = new Date(now);
-    monday.setDate(now.getDate() + diff);
-    monday.setHours(0, 0, 0, 0);
+    monday.setUTCDate(now.getUTCDate() + diff);
+    monday.setUTCHours(0, 0, 0, 0);
     return monday.toISOString();
   }
   if (recurrence === "monthly") {
-    return new Date(now.getFullYear(), now.getMonth(), 1, 0, 0, 0, 0).toISOString();
+    // Date.UTC avoids the local-timezone offset that the Date constructor
+    // applies when month/day/hour arguments are passed directly.
+    return new Date(Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), 1)).toISOString();
   }
   return new Date(0).toISOString(); // 'none' never resets
 }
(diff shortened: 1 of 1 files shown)


Priyanshu-byte-coder (approved):
Both fixes are correct. The optimistic lock (`.lt('period_start', ...)`) ensures only one concurrent reset wins — the losing request re-fetches the already-reset row instead of silently zeroing progress. The UTC boundary fix (`getUTCDay`/`setUTCHours`) prevents server-timezone drift from shifting Monday to Sunday or Tuesday.

shaikarshiya451-beep:
fix(bugs):correct the bugs

PR #347, opened 2026-05-19
fix(streak): paginate GitHub commit search to avoid 100-result truncation

### LINKED ISSUE #346: [Bug]: Commit streak truncated for users with more than 100 commits in the 90-day window
## Description

`fetchActiveDates` in `src/app/api/metrics/streak/route.ts` fetches only the **first page** (100 results) of the GitHub Search API. Active users who commit more than 100 times within 90 days have their oldest commits silently discarded, which introduces phantom gaps in the active-date set and causes the streak calculator to report a shorter (incorrect) streak.

## Current behavior

1. User commits every day for 90 days (90+ commits total, many days with multiple commits).
2. API fetches only the 100 most recent commits.
3. Days covered only by commits beyond the 100th are absent from `activeDates`.
4. `calculateStreakFromDates` finds a gap at those missing days and resets the streak counter.
5. Dashboard shows a streak shorter than the user actually has.

`totalActiveDays` and `longestStreak` are also undercounted for the same reason.

## Root cause

```typescript
// src/app/api/metrics/streak/route.ts, line 29
const searchRes = await fetch(
  `${GITHUB_API}/search/commits?q=author:${githubLogin}+author-date:>=${sinceStr}&per_page=100&sort=author-date&order=desc`,
  // ...
);
// Only the first 100 commits are ever processed -- no pagination
```

## Fix

Loop through pages until the response contains fewer than 100 items (indicating the last page), or page 10 is reached (GitHub Search API hard cap of 1000 results):

```typescript
let page = 1;
while (true) {
  const searchRes = await fetch(
    `${GITHUB_API}/search/commits?q=author:${githubLogin}+author-date:>=${sinceStr}&per_page=100&page=${page}&sort=author-date&order=desc`,
    { headers: { ... }, cache: "no-store" }
  );
  // ... process items ...
  if (data.items.length < 100 || page >= 10) break;
  page++;
}
```

## Who is affected

Any user who has made more than 100 commits across all tracked repositories within the last 90 days. Prolific open-source contributors and daily committers are most impacted.

/assign gssoc

anshul23102:
Fixes #346

## What changed

`fetchActiveDates` in `src/app/api/metrics/streak/route.ts` now paginates through the GitHub Search API instead of stopping after the first 100 results.

The single `fetch` call is replaced with a `while` loop that:
- Adds each page's commits to the `activeDates` set
- Breaks when the page returns fewer than 100 items (last page)
- Breaks at page 10 (GitHub Search API hard cap of 1000 results)

No other logic in `calculateStreakFromDates` or the route handler is changed.

## Why

GitHub Search API returns at most 100 results per page. Users with more than 100 commits in the 90-day window had their oldest commits silently dropped, introducing phantom gaps that shortened the reported streak. `totalActiveDays` and `longestStreak` were also undercounted.

## Testing

1. Use a GitHub account with 100+ commits in the last 90 days.
2. Before: streak shows a gap or resets unexpectedly.
3. After: streak reflects the continuous commit history correctly.
4. Users with fewer than 100 commits in the window see no change in behavior (loop exits on first iteration).

---
> Draft -- waiting for assignment on #346 before review is requested.

### DIFF
--- a/src/app/api/goals/route.ts
+++ b/src/app/api/goals/route.ts
@@ -9,15 +9,21 @@ type Recurrence = "none" | "weekly" | "monthly";
 function getPeriodStart(recurrence: Recurrence): string {
   const now = new Date();
   if (recurrence === "weekly") {
-    const day = now.getDay();
+    // Use UTC methods so the Monday boundary is the same regardless of the
+    // server's local timezone. getDay() / setDate() / setHours() all operate
+    // in local time, which can push the reset boundary a day early or late
+    // on servers that are not running in UTC.
+    const day = now.getUTCDay();
     const diff = day === 0 ? -6 : 1 - day; // Monday
     const monday = new Date(now);
-    monday.setDate(now.getDate() + diff);
-    monday.setHours(0, 0, 0, 0);
+    monday.setUTCDate(now.getUTCDate() + diff);
+    monday.setUTCHours(0, 0, 0, 0);
     return monday.toISOString();
   }
   if (recurrence === "monthly") {
-    return new Date(now.getFullYear(), now.getMonth(), 1, 0, 0, 0, 0).toISOString();
+    // Date.UTC avoids the local-timezone offset that the Date constructor
+    // applies when month/day/hour arguments are passed directly.
+    return new Date(Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), 1)).toISOString();
   }
   return new Date(0).toISOString(); // 'none' never resets
 }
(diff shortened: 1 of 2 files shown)


Priyanshu-byte-coder (approved):
Pagination correct. The while loop stops when `page * per_page >= total_count` (all commits fetched) or when an empty page is returned — handles the edge case where total_count is stale. Same UTC-boundary fix as #345 applied to the streak route. Prevents active users with >100 commits in the period from having their streak miscalculated.

Issue #301, opened 2026-05-17
[BUG] Paginate commit search results for high-activity users


drb101005:
## Bug: Commit search only fetches the first page of results

While looking into some inaccurate contribution metrics, I noticed that `fetchCommitSearch` in `route.ts` (around line 53) only makes a single GitHub Search API request with `per_page=100`.

The problem is that GitHub search results are paginated, so for users with more than 100 matching commits we're only processing the first page of data. This causes several metrics to be undercounted or skewed, including:

- total commits
- most active repo
- active contribution days
- contribution deltas/trends

### Proposed fix

Update `fetchCommitSearch` to paginate through all available search results instead of stopping after the first request.

Suggested approach:

- Request results with:
  - `per_page=100`
  - `page=N`
- Keep fetching pages until:
  - fewer than 100 items are returned
  - we've reached `total_count`
  - or we hit GitHub Search API's 1000-result cap
- Aggregate all returned items before returning the final response

Expected return shape should still look like:

```ts
{
  total_count,
  items
}

drb101005:
Hey @Priyanshu-byte-coder 
Can i be assigned this Issue under GSSoC , Also let me know if you have any suggestions for the proposed fix

Nikhithapriya2548:
Hi @Priyanshu-byte-coder! I am a GSSoC '26 contributor and I would love to work on fixing this low contrast text issue on the landing page to improve UI/UX and accessibility. 

My Fix Plan:
1. I will locate the landing page styling file and find the text elements with poor visibility.
2. I will update the color contrast ratios (text color vs background) to meet standard accessibility guidelines.
3. I'll test the responsiveness and readability across different screen layouts before opening a clean PR.

Could you please assign this issue to me? Ready to start right away!

Jeshika311:
Hi @Priyanshu-byte-coder 

I would like to work on this issue under GSSoC'26.

I understood that `fetchCommitSearch` currently fetches only the first page of GitHub Search API results, which causes inaccurate contribution metrics for users having more than 100 commits.

I can fix this by implementing proper pagination using:

* `per_page=100`
* `page=n`

and aggregating all fetched results while handling GitHub API limits and maintaining the existing response structure.

I’ll also test the fix properly to ensure accurate commit counts and contribution metrics.

Please assign this issue to me. Thank you!

pithva007:
Hi @drb101005 👋

I would like to work on this issue under GSSoC'26.

[...]

PR #311, opened 2026-05-18
feat: cache GitHub metrics with Upstash Redis

saurabhhhcodes:
## Summary
- add a reusable Upstash Redis metrics cache helper with graceful no-env fallback
- cache GitHub-backed metrics for contributions, repos, PR summaries, and streak active dates
- apply endpoint-specific TTLs: contributions 5m, repos 10m, PRs 10m, streak 2m
- add deterministic keys scoped by user/account, endpoint, and query params
- support explicit cache bypass/refresh via `refresh=1`, `bypassCache=true`, `sync=true`, or `x-devtrack-cache-bypass: 1`
- document optional Upstash env vars in `.env.example`

Fixes #234

## GSSoC label request
If accepted, please add:
- `gssoc:approved`
- `level:advanced`
- `quality:clean`
- `type:performance`
- `type:backend`

## Testing
- `node /tmp/codex-npm-cli/bin/npm-cli.js run type-check` ✅
- `node /tmp/codex-npm-cli/bin/npm-cli.js run lint` ✅ passes with pre-existing warnings in `BadgeSection.tsx` and `CommitTimeChart.tsx`
- `git diff --check` ✅
- `node -e "JSON.parse(require('fs').readFileSync('package-lock.json','utf8'))"` ✅

## Build note
- `node /tmp/codex-npm-cli/bin/npm-cli.js run build` compiles successfully, then fails while collecting page data because this local environment does not have Supabase env configured (`supabaseUrl is required`) in `/api/auth/link-github`. That is unrelated to the metrics cache change.

PR #399, opened 2026-05-19
Fix commit search pagination in contributions metrics

drb101005:
Fixed inaccurate contribution metrics in route.ts by properly paginating GitHub commit search results
Added support for fetching multiple pages (per_page=100&page=N) until all available results are retrieved, the reported total count is reached, or GitHub’s 1000-result search limit is hit
Aggregated commit data across all pages before calculating contribution statistics
Improved the accuracy of total commits, active contribution days, and related contribution metrics while keeping the existing API response shape unchanged

Closes [BUG] Paginate commit search results for high-activity users #301

### DIFF
--- a/src/app/api/metrics/contributions/route.ts
+++ b/src/app/api/metrics/contributions/route.ts
@@ -65,6 +65,9 @@ async function fetchContributionsForAccount(
       let totalCount = 0;
       let page = 1;
 
+      // Note: this may issue up to 10 sequential GitHub Search API calls (max 1000 results).
+      // Authenticated GitHub Search rate limits are low (~30 req/min). We handle 429/403
+      // responses gracefully by returning partial results rather than failing the endpoint.
       while (page <= 10) {
         const searchRes = await fetch(
           `${GITHUB_API}/search/commits?q=author:${githubLogin}+author-date:>=${sinceStr}&per_page=100&page=${page}&sort=author-date&order=desc`,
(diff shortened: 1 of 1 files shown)


drb101005 (commented):
Seen and made the changes

Priyanshu-byte-coder (changes requested):
The pagination logic is correct, but this PR removes the `withMetricsCache` wrapper that was added by the recently merged PR #311. Removing it breaks the Redis caching for contributions.

The pagination needs to be implemented INSIDE the `withMetricsCache` callback:

```ts
return withMetricsCache(
  { bypass: cacheContext.bypass, key, ttlSeconds: METRICS_CACHE_TTL_SECONDS.contributions },
  async () => {
    const since = new Date();
    since.setDate(since.getDate() - days);
    const sinceStr = toLocalDateStr(since);

    let allItems: Array<{ commit: { author: { date: string } } }> = [];
    let page = 1;

    while (page <= 10) {
      // ... pagination loop ...
      if (data.items.length < 100) break;
      page++;
    }

    const commitsByDay: Record<string, number> = {};
    for (const item of allItems) { ... }

    return { days, total: totalCount, data: commitsByDay };
  }
);
```

Please rebase on `main` (which has #311 merged) and add pagination inside the cache callback.

Priyanshu-byte-coder (changes requested):
**Pagination logic is correct** — the loop up to 10 pages with the 1000-result cap is the right approach.

**Rate limit risk** — up to 10 sequential GitHub Search API calls per load for high-activity users. GitHub's authenticated rate limit is 30 req/min for search. At minimum add graceful handling for 429/403 responses (currently throws a generic error). Ideally add a comment acknowledging this trade-off so future maintainers understand the risk.

Priyanshu-byte-coder (changes requested):
This PR has merge conflicts with main. Please rebase on the latest main branch and force-push to resolve:

```
git fetch origin
git rebase origin/main
git push --force-with-lease
```

PR #410, opened 2026-05-19
fix(TopRepos): show repo description as native tooltip on name hover (#214)

### LINKED ISSUE #214: [GOOD FIRST ISSUE] Show repo description as tooltip on name hover in TopRepos
## Summary
TopRepos lists repo names but not descriptions, even though the GitHub API returns them. Useful context is hidden.

## Expected Behaviour
Hover over repo name shows description as a tooltip (if available).

## Implementation Notes
- Verify description field is passed through from API to component
- Add title={repo.description || undefined} to repo name element
- If description is null/empty, show nothing

## GSSoC Info
- Difficulty: Beginner — 20 pts
- Area: UI / DX

Jyoti1-pog:
## Summary

Fixes #214 — `TopRepos` listed repository names but never surfaced the
description the GitHub API already returns. Hovering a repo name now shows
its description as a native browser tooltip. Repos with no description
(`null` or `""`) correctly show no tooltip at all.

## Changes

- **`src/components/TopRepos.tsx`** — added `description: string | null` to
  the `Repo` type; set the repo name link's
  `title={repo.description || undefined}`.
- **`src/app/api/metrics/repos/route.ts`** — threaded `description` through
  the GitHub search response type, the `repoMap` projection, the
  `RepoSummary` type, and the multi-account `mergeRepoCommits` merge. The
  field was being dropped at every step of the API → component pipeline.

Using `|| undefined` (not `""`) makes React omit the `title` attribute
entirely when there is no description, so no empty tooltip renders.

## Verification

Confirmed against the live GitHub API that `search/commits` returns
`repository.description`, and verified end-to-end on the dashboard that a
repo with a description shows the tooltip while a repo with a `null`
description shows none.

**Tooltip shows on hover (repo with a description):**

Screenshot 2026-05-20 021225

**No tooltip (repo with null/empty description):**

Screenshot 2026-05-20 021557

## Note

The repo name link previously had `title={repo.name}` (full `org/repo`
path on hover). An element can carry only one `title`, and the issue
explicitly specifies `title={repo.description || undefined}` on the name
element, so that prior tooltip is now replaced by the description — this is
the behaviour the issue asks for.

Closes #214

### DIFF
--- a/src/app/api/metrics/repos/route.ts
+++ b/src/app/api/metrics/repos/route.ts
@@ -21,6 +21,7 @@ export const dynamic = "force-dynamic";
 interface RepoSummary {
   name: string;
   commits: number;
+  description: string | null;
 }
 
 interface RepoResponse {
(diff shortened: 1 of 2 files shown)


Priyanshu-byte-coder (changes requested):
**Description merge logic prefers `null` over real values** — in `mergeRepoCommits`, `existing?.description ?? repo.description` keeps `null` if the first encounter for a repo had no description, even if a later encounter does. Change to prefer non-null:

```ts
description: existing.description ?? repo.description,
```

This is already correct — `??\ falls through null/undefined so it will prefer `repo.description` when `existing.description` is null. Actually the issue is reversed: the current code is `existing?.description ?? repo.description` which means: if existing exists and has a description use it, else use repo.description. But if `existing.description` is `null`, `null ?? repo.description` gives `repo.description` — this is actually correct.

Re-reviewing: the logic is fine. The core feature (threading description through to the `title` attribute) is correct and complete. **APPROVE** — this is mergeable.

Priyanshu-byte-coder (approved):
Description merge logic (`existing?.description ?? repo.description`) correctly prefers non-null via `??`. Feature is complete and correct — threads `description` from API through to the `title` attribute with proper `|| undefined` for empty strings. Merging.

PR #441, opened 2026-05-20
fix : added XML escaping for badge SVG accessible labels

### LINKED ISSUE #440: fix : add XML escaping for badge SVG accessible labels
Badge SVG generation escapes visible text nodes, but the aria-label and title content are built from raw label and value strings. The badge generator should use escaped text consistently so repository names, labels, or metric values containing XML-sensitive characters cannot produce malformed SVG output.

Changes that Need to be Made:
Update src/app/api/badge/badge-utils.ts so generateBadgeSVG computes an escaped label and escaped value once, then uses those escaped strings in aria-label, title, and text elements. Keep the generated layout and colors unchanged.

Impact that it would Provide:
This improves badge reliability and accessibility metadata correctness for labels and values containing characters such as &, <, >, quotes, or apostrophes.

tmdeveloper007:
Closes #440.

Badge SVG generation now escapes label and value text before inserting those strings into accessible metadata and visible SVG text nodes.

Changes Made:
Updated src/app/api/badge/badge-utils.ts to compute escapedLabel and escapedValue once in generateBadgeSVG, then reuse them for aria-label, title, shadow text, and foreground text.

Impact it Made:
Badges remain valid when labels or values contain XML-sensitive characters such as &, <, >, quotes, or apostrophes. Verification completed with npm run type-check.

### DIFF
--- a/src/app/api/badge/badge-utils.ts
+++ b/src/app/api/badge/badge-utils.ts
@@ -19,6 +19,8 @@ export function generateBadgeSVG({
   color = "#6366f1", // DevTrack accent color (indigo)
   labelColor = "#333333",
 }: BadgeConfig): string {
+  const escapedLabel = escapeXml(label);
+  const escapedValue = escapeXml(value);
   // SVG dimensions
   const labelWidth = label.length * 7 + 10;
   const valueWidth = value.length * 8 + 10;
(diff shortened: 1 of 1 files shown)


tmdeveloper007:
@Priyanshu-byte-coder 
Review requestred.

PR #443, opened 2026-05-20
fix: return null instead of throwing on invalid crypto input

### LINKED ISSUE #442: fix : add validation for encrypted token payload lengths
Token decryption should reject malformed encrypted payloads and IV values before constructing the AES-GCM decipher. The current implementation accepts any hex string shape and lets low-level crypto errors surface.

Changes that Need to be Made:
Update src/lib/crypto.ts so decryptToken validates that the IV is exactly 12 bytes encoded as 24 hex characters and that the encrypted payload contains at least the 16-byte authentication tag encoded as hex. Reuse the existing error style with clear messages.

Impact that it would Provide:
This makes token decryption failures easier to diagnose and avoids passing invalid buffer sizes into crypto primitives, improving reliability around linked GitHub account token handling.

tmdeveloper007:
Closes #442.

Updated  to return  instead of throwing on invalid input, preserving the original return type contract. Also fixed the edge case where empty ciphertext could pass validation by changing  to  in the length check.

Changes Made:
- File: src/lib/crypto.ts
  - Changed return type from  to 
  - Validation failures now return  instead of throwing errors
  - Changed  to 

- File: src/lib/github-accounts.ts
  - Added filter to remove null tokens in 
  - Added filter to remove invalid accounts in 

Impact it Made:
- All call sites continue to work without modification
- Invalid token data returns null consistently across the codebase

### DIFF
--- a/src/lib/crypto.ts
+++ b/src/lib/crypto.ts
@@ -9,6 +9,9 @@ const IV_LENGTH = 12;
 const AUTH_TAG_LENGTH = 16;
 const KEY_ERROR_MESSAGE =
   "ENCRYPTION_KEY env var must be a 32-byte hex string";
+const ENCRYPTED_TOKEN_ERROR_MESSAGE =
+  "Encrypted token must be a hex string containing ciphertext and auth tag";
+const IV_ERROR_MESSAGE = "Token IV must be a 12-byte hex string";
 
 function getEncryptionKey(): Buffer {
   const key = process.env.ENCRYPTION_KEY;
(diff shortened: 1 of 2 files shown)


tmdeveloper007:
@Priyanshu-byte-coder 
Review requestred.

Priyanshu-byte-coder (changes requested):
Two issues to resolve before merge:

1. **Return type contract changed without auditing callers.** The base `decryptToken` returns `string | null` — all call sites in `github-accounts.ts` handle `null` correctly. This PR appears to change the return type to `string` and throws instead of returning `null`. If the existing `try/catch` wrapper is still present, the thrown errors are silently caught and the validation is dead code. If the `try/catch` was removed, all callers now receive uncaught exceptions instead of null. Please clarify: keep returning `null` on invalid input (consistent with existing contract) rather than throwing, so no call sites need to change.

2. **Length check edge case.** `encrypted.length < AUTH_TAG_LENGTH * 2` (i.e., < 32) allows zero-byte ciphertext (length === 32 = auth tag only, empty plaintext). If empty plaintext tokens should be rejected, use `<= AUTH_TAG_LENGTH * 2`.

Priyanshu-byte-coder (changes requested):
Two issues still not resolved + new merge conflict:

1. **Merge conflict** — branch needs rebase on `main` before this can proceed.
2. **Return type not fixed** — `decryptToken` still throws instead of returning `null`. Callers in `github-accounts.ts` expect `null` on bad input; they will now receive uncaught exceptions. Keep `string | null` return type and return `null` from the guard clauses instead of throwing.
3. **Length check** — `encrypted.length < AUTH_TAG_LENGTH * 2` allows zero-byte ciphertext (32 hex chars = 16-byte auth tag + 0 bytes plaintext). Use `<= AUTH_TAG_LENGTH * 2` to reject empty plaintext.

tmdeveloper007 (commented):
Thank you for the review. I have addressed both issues:

1. **Return type contract preserved**: Changed  to return  instead of throwing. Added filters in  and  to handle null values gracefully.

2. **Length check edge case fixed**: Changed  to  to reject empty ciphertext.

All tests pass (type-check and lint). Please review the updated changes.

PR #445, opened 2026-05-20
fix : added flexible metrics cache bypass parsing

### LINKED ISSUE #444: fix : add flexible metrics cache bypass parsing
Metrics cache bypass handling currently recognizes only exact refresh or bypass values of 1 and true for query parameters, and only 1 for the x-devtrack-cache-bypass header. This should be normalized so common truthy values work consistently.

Changes that Need to be Made:
Update src/lib/metrics-cache.ts to introduce a small helper for cache bypass values. It should trim and lowercase input values, accept 1, true, yes, and on, and apply the same parsing to query parameters and the x-devtrack-cache-bypass header.

Impact that it would Provide:
This improves developer and API ergonomics for manual metric refreshes while keeping cache behavior unchanged for default requests.

tmdeveloper007:
Closes #444.

Metrics cache bypass parsing now normalizes query parameter and header values before deciding whether to skip cached metric responses.

Changes Made:
Updated src/lib/metrics-cache.ts with an isTruthyCacheBypass helper that trims and lowercases values, then accepts 1, true, yes, and on for both query parameters and the x-devtrack-cache-bypass header.

Impact it Made:
Manual metric refreshes are more forgiving and consistent across request surfaces without changing default cached behavior. Verification completed with npm run type-check.

### DIFF
--- a/src/lib/metrics-cache.ts
+++ b/src/lib/metrics-cache.ts
@@ -13,6 +13,14 @@ type CacheParamValue = boolean | number | string | null | undefined;
 
 let redisClient: Redis | null | undefined;
 
+function isTruthyCacheBypass(value: string | null): boolean {
+  if (!value) {
+    return false;
+  }
+
+  return ["1", "true", "yes", "on"].includes(value.trim().toLowerCase());
+}
+
 function getRedisClient(): Redis | null {
   if (redisClient !== undefined) {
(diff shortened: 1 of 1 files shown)


tmdeveloper007:
@Priyanshu-byte-coder 
Review requestred.

PR #447, opened 2026-05-20
fix : added finite number handling to repository health scoring

### LINKED ISSUE #446: fix : add finite number handling to repository health scoring
Repository health scoring should defend against non-finite metric values such as NaN, Infinity, or -Infinity. The current clamp helper can return NaN when given NaN, which can lead to invalid score output.

Changes that Need to be Made:
Update src/lib/repo-health.ts so numeric clamping coerces non-finite values to a safe fallback before applying min and max bounds. Keep the existing scoring weights and grade thresholds unchanged.

Impact that it would Provide:
This makes repository health responses more robust when upstream GitHub metric calculations are missing, divided by zero, or otherwise non-finite, ensuring the API still returns a bounded numeric score.

tmdeveloper007:
Closes #446.

Repository health scoring now handles non-finite numeric inputs before applying score bounds.

Changes Made:
Updated src/lib/repo-health.ts so the shared clamp helper returns the minimum bound when it receives NaN, Infinity, or -Infinity, while preserving the existing scoring weights and grade thresholds.

Impact it Made:
Health score responses stay numeric and bounded even when upstream metric calculations produce non-finite values. Verification completed with npm run type-check.

### DIFF
--- a/src/lib/repo-health.ts
+++ b/src/lib/repo-health.ts
@@ -1,6 +1,10 @@
 import type { RepoHealthScore, RepoHealthSignals } from "@/types/repo-health";
 
 function clamp(n: number, min: number, max: number): number {
+  if (!Number.isFinite(n)) {
+    return min;
+  }
+
   return Math.min(max, Math.max(min, n));
 }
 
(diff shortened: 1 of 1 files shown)


tmdeveloper007:
@Priyanshu-byte-coder 
Review requestred.

PR #449, opened 2026-05-20
fix : added side-effect import detection to dependency checker

### LINKED ISSUE #448: fix : add side-effect import detection to dependency checker
The dependency checker scans normal import and export-from statements, but it does not detect side-effect imports such as import "package";. This can let undeclared packages pass the check.

Changes that Need to be Made:
Update scripts/check-deps.js to parse side-effect imports in addition to existing static and dynamic imports. Export small parser helpers behind a CommonJS module interface and add a node:test suite using node:test and node:assert/strict to verify scoped package extraction, dependency detection, and side-effect import coverage.

Impact that it would Provide:
This strengthens dependency validation and adds regression coverage for the checker so future import syntax changes are less likely to bypass package.json validation.

tmdeveloper007:
Closes #448.

The dependency checker now detects side-effect imports and has focused regression tests for import parsing and missing dependency detection.

Changes Made:
Updated scripts/check-deps.js to parse side-effect imports, allow indented static imports, expose parser helpers for tests, and preserve the same command-line behavior through a main guard. Added test/check-deps.test.js using node:test and node:assert/strict to cover scoped package extraction, side-effect import extraction, and missing dependency reporting.

Impact it Made:
Undeclared packages imported only for side effects are now caught by the dependency check, and the parser behavior has local test coverage. Verification completed with node --test test/check-deps.test.js, node scripts/check-deps.js, and npm run type-check.

### DIFF
--- a/scripts/check-deps.js
+++ b/scripts/check-deps.js
@@ -8,12 +8,6 @@
 const fs = require("fs");
 const path = require("path");
 
-const pkg = JSON.parse(fs.readFileSync("package.json", "utf8"));
-const allDeps = new Set([
-  ...Object.keys(pkg.dependencies || {}),
-  ...Object.keys(pkg.devDependencies || {}),
-]);
-
 // Node built-ins (not in package.json but valid to import)
 const BUILTINS = new Set([
   "assert", "buffer", "child_process", "cluster", "crypto", "dgram",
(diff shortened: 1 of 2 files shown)


tmdeveloper007:
@Priyanshu-byte-coder 
Review requestred.

PR #496, opened 2026-05-21
fix : added node:test and node:assert/strict to BUILTINS in check-deps

tmdeveloper007:
This PR fixes the check-deps script to recognize node:test and node:assert/strict as Node.js built-in modules. This allows test files using Node's native test module to pass CI.


Priyanshu-byte-coder:
Closing — superseded by #497 which is the complete fix. It handles both prefixed (`node:test`) and non-prefixed imports correctly via a dedicated `NODE_BUILTINS` set, whereas this PR only covers non-prefixed imports.

PR #508, opened 2026-05-21
fix : validate repository_url before extracting repo name

### LINKED ISSUE #498: fix : validate repository_url before extracting repo name in github.ts
The code in src/lib/github.ts assumes repository_url always contains a properly formatted GitHub API URL. If the URL is malformed, split("/").pop() could return undefined or an empty string.

Changes that Need to be Made:
- File: src/lib/github.ts (lines 113-118)
- Add validation to ensure URL has expected structure before extracting repo name
- Handle cases where URL might be malformed

Impact that it would Provide:
- Prevents crashes from malformed URLs
- Makes the code more robust against unexpected API responses

tmdeveloper007:
Closes #498.

Added validation to ensure repository_url has the expected structure before extracting the repo name. Previously, malformed URLs would result in empty string keys in the repoCounts map.

Changes Made:
- File: src/lib/github.ts
- Check that URL has at least 2 path segments before extracting repo name
- Only increment count if repo name is non-empty

Impact it Made:
- Prevents crashes from malformed URLs
- Makes the code more robust against unexpected API responses

### DIFF
--- a/src/lib/github.ts
+++ b/src/lib/github.ts
@@ -110,8 +110,11 @@ export async function fetchIssuesMetrics(
 
   const repoCounts: Record<string, number> = {};
   for (const item of items) {
-    const repo = item.repository_url.split("/").pop() ?? "";
-    repoCounts[repo] = (repoCounts[repo] ?? 0) + 1;
+    const parts = item.repository_url.split("/");
+    const repo = parts.length >= 2 ? parts.pop()! : item.repository_url;
+    if (repo) {
+      repoCounts[repo] = (repoCounts[repo] ?? 0) + 1;
+    }
   }
   const mostActiveRepo =
     Object.keys(repoCounts).length > 0


Priyanshu-byte-coder (changes requested):
Fallback branch is wrong. When `parts.length < 2`, the code falls back to `item.repository_url` (the full raw URL string) as the repo key. The `if (repo)` guard below treats any non-empty string as valid, so the entire URL like `https://api.github.com/repos/owner/name` gets counted as a repo name.

Simpler fix with no edge cases:
```ts
const repo = item.repository_url.split("/").at(-1) ?? "";
if (repo) { repoCounts[repo] = (repoCounts[repo] ?? 0) + 1; }
```
`Array.at(-1)` always returns the last segment or `undefined` — no length guard needed.

PR #593, opened 2026-05-21
fix: remove Playwright auth bypass from dashboard (GHSA-6c5j-4w43-2v8f #1)

advikdivekar:
## Problem

The dashboard page contained a cookie-based authentication bypass that could be activated in production by any visitor. Two conditions enabled it: the environment variable `PLAYWRIGHT_AUTH_BYPASS=1` and a browser cookie `playwright-dashboard-auth=1` that was neither server-set nor HttpOnly.

Any contributor who copied the `playwright.config.mjs` env block into their Vercel deployment activated this bypass. From there, anyone who knew about it could set the cookie via DevTools and access `/dashboard` without GitHub OAuth.

**Root cause:** Production code path that sets `session = null` and skips `getServerSession`, controlled by a user-settable cookie.

## What changed

**`src/app/dashboard/page.tsx`**
- Removed the `allowPlaywrightBypass` condition and the `cookies()` import entirely
- Authentication is now enforced unconditionally via `getServerSession(authOptions)`
- E2e tests that require an authenticated session should use proper NextAuth JWT injection in the test setup rather than a production code path

## How to verify

1. Deploy with `PLAYWRIGHT_AUTH_BYPASS=1` set in env vars
2. Open browser DevTools → run `document.cookie = "playwright-dashboard-auth=1"`
3. Navigate to `/dashboard` — redirected to `/` instead of rendering the dashboard
4. Normal sign-in flow: unaffected

## Regression check

- Authenticated users: dashboard loads normally
- Unauthenticated users: redirected to `/` as before
- No other references to `PLAYWRIGHT_AUTH_BYPASS` or `playwright-dashboard-auth` in application code

Fixes GHSA-6c5j-4w43-2v8f vulnerability #1 (Critical).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,48 @@
+# -------------------------------------------------------
+# Supabase
+# Project Settings → API → Project URL
+NEXT_PUBLIC_SUPABASE_URL=https://your-project-ref.supabase.co
+
+# Project Settings → API → anon / public key
+NEXT_PUBLIC_SUPABASE_ANON_KEY=your_supabase_anon_key
+
+# Project Settings → API → service_role secret (server-side only — never expose client-side)
+SUPABASE_SERVICE_ROLE_KEY=your_supabase_service_role_key
+
+# -------------------------------------------------------
+# NextAuth
+# The URL your app is running on
+NEXTAUTH_URL=http://localhost:3000
+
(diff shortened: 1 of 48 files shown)


advikdivekar:

PR #594, opened 2026-05-21
fix: re-validate GitHub OAuth token every 24h to detect revocation (GHSA-6c5j-4w43-2v8f #2)

advikdivekar:
## Problem

The NextAuth `jwt` callback stored the GitHub OAuth access token at sign-in and never re-validated it. The `account` object is only populated on the first sign-in event — every subsequent JWT refresh skipped that block entirely. The 30-day JWT continued to hold and submit a revoked token for its full lifetime with no detection.

A user who revoked DevTrack's access via GitHub Settings had no way to actually invalidate the session — DevTrack kept appearing authenticated and kept submitting the revoked token as `Authorization: Bearer` on every API call for up to 30 days.

**Root cause:** No periodic token validation in the `jwt` callback. The `error` field was absent from type declarations, so there was no type-safe path to signal revocation downstream.

## What changed

**`src/lib/auth.ts`**
- On initial sign-in, stamps `token.accessTokenValidatedAt = Date.now()` alongside the access token
- On every subsequent JWT callback: if more than 24 hours have elapsed since last validation, makes a single lightweight `GET /user` call to the GitHub API
  - `401` response → sets `token.error = "TokenRevoked"`
  - `200` response → updates `token.accessTokenValidatedAt`, clears any prior error
  - Network error or GitHub 5xx → session left intact (no false evictions)
- `session` callback now surfaces `token.error` into `session.error`

**`src/types/next-auth.d.ts`**
- Added `error?: string` to the `Session` interface
- Added `accessTokenValidatedAt?: number` and `error?: string` to the `JWT` interface

**`src/app/dashboard/page.tsx`**
- After the null-session check, redirects to `/?error=TokenRevoked` when `session.error === "TokenRevoked"`

## How to verify

1. Sign in to DevTrack with a GitHub account
2. Go to **GitHub Settings → Applications → Authorized OAuth Apps → DevTrack → Revoke**
3. Return to `/dashboard` without signing out — within 24 hours the page redirects to `/?error=TokenRevoked`
4. TypeScript: `session.error` compiles without type error across the codebase
5. Network failure during token check: session preserved, no redirect

## Regression check

- Normal sign-in and dashboard use: unaffected (validation triggers only after 24h)
- All API routes reading `session.accessToken`: unchanged
- `SESSION_MAX_AGE` and `SESSION_UPDATE_AGE`: unchanged

Fixes GHSA-6c5j-4w43-2v8f vulnerability #2 (High).


advikdivekar:

PR #595, opened 2026-05-21
fix: isolate compare route cache per user to prevent cross-user data exposure (GHSA-6c5j-4w43-2v8f #3)

advikdivekar:
## Problem

The `/api/metrics/compare` route used `next: { revalidate: 3600 }` on four GitHub API fetch calls. Next.js's built-in fetch cache keys responses by URL only — the `Authorization` header is not part of the cache key.

This meant: if User A (with private repositories) queried a GitHub username, their response was cached under the GitHub API URL. User B querying the same username within the one-hour window received User A's cached response, which could include commit counts and repository metadata from User A's private repositories.

**Root cause:** `next: { revalidate }` is incompatible with per-user authenticated requests. Any user's private data fetched with their token becomes visible to any other user who queries the same target URL.

## What changed

**`src/app/api/metrics/compare/route.ts`**
- Replaced all four `next: { revalidate: 3600 }` options with `cache: "no-store"` to opt out of the shared URL-keyed fetch cache
- Wrapped the full data-fetch and computation block in `withMetricsCache` (existing utility from `@/lib/metrics-cache`) with a cache key of `metrics:{session.githubId}:compare:{username}`
- The `githubId` component ensures no two users ever share a cache entry
- Added `isMetricsCacheBypassed` support so clients can force a cache refresh with `?refresh=1`
- Gracefully degrades to live fetches when Redis is unavailable

## How to verify

1. Sign in as User A (account with private repositories), use the Friend Comparison widget to look up any GitHub username
2. Sign in as User B in a separate session, look up the same username within one hour
3. User B sees only data their token can access — no private repo data from User A leaks through
4. `?refresh=1` query param bypasses the cache and fetches fresh data

## Regression check

- Friend Comparison widget: still returns correct data
- Cache degrades gracefully when `UPSTASH_REDIS_REST_URL` / `UPSTASH_REDIS_REST_TOKEN` are not set
- Auth check and username validation: unchanged

Fixes GHSA-6c5j-4w43-2v8f vulnerability #3 (High).

### DIFF
--- a/src/app/api/metrics/compare/route.ts
+++ b/src/app/api/metrics/compare/route.ts
@@ -1,6 +1,10 @@
 import { getServerSession } from "next-auth";
 import { NextRequest } from "next/server";
 import { authOptions } from "@/lib/auth";
+import {
+  isMetricsCacheBypassed,
+  withMetricsCache,
+} from "@/lib/metrics-cache";
 
 export const dynamic = "force-dynamic";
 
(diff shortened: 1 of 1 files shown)


advikdivekar:

PR #596, opened 2026-05-21
fix: harden badge endpoints against rate limit exhaustion and username disclosure (GHSA-6c5j-4w43-2v8f #4)

advikdivekar:
## Problem

Both badge endpoints (`/api/badge/commits` and `/api/badge/streak-shield`) were fully unauthenticated, accepted arbitrary `?user=` values with only a 50-character length check, and made direct GitHub API calls consuming the shared `GITHUB_TOKEN` quota with no rate limiting.

Three concrete issues:

1. **Rate limit exhaustion**: An automated script looping over badge endpoints with no auth requirement could exhaust the 5,000 req/hr `GITHUB_TOKEN` quota in ~83 minutes, taking down the leaderboard, public profiles, and all badge features for every user until the window reset.

2. **Username disclosure via logs**: Both routes called `console.log` for every request — every queried username and its commit/streak count was written to Vercel function logs, creating a persistent record accessible to anyone with log access.

3. **Unvalidated username characters**: The only check was a 50-char length limit. Characters like `/`, `?`, `+`, and `#` were passed directly into GitHub API URL path and query segments.

**Root cause:** Badge routes were excluded from the existing rate limiting middleware (matcher only covered `/api/metrics/*`) and had no format validation.

## What changed

**`src/middleware.ts`**
- Extended `matcher` to `["/api/metrics/:path*", "/api/badge/:path*"]`
- Badge routes are always unauthenticated, so the middleware applies `ANONYMOUS_LIMIT` (10 req/min per IP) — no other middleware changes needed

**`src/app/api/badge/commits/route.ts`** and **`src/app/api/badge/streak-shield/route.ts`**
- Replaced the weak 50-char length check with a strict GitHub username regex: alphanumeric and hyphens, no leading/trailing hyphens, no consecutive hyphens, max 39 characters
- Removed all `console.log` statements that disclosed queried usernames and their data
- Added per-username response caching using `cacheGet`/`cacheSet` from `@/lib/metrics-cache` (TTL 3600s) — repeated requests for the same username within the cache window skip the GitHub API call entirely
- Cache keys are `badge:commits:{username}` and `badge:streak:{username}` — public data, no per-user scoping needed
- Gracefully falls through to live fetches when Redis is unavailable

## How to verify

1. `curl "/api/badge/commits?user=valid-user"` → 200 SVG response
2. `curl "/api/badge/commits?user=invalid/user"` → 400 JSON error
3. Send 11 rapid requests from the same IP → 11th returns 429 with `Retry-After` header
4. Check Vercel function logs — no username appears in log output
5. Two requests for the same username within 3600s → second request served from cache (no additional GitHub API call)

## Regression check

- Valid badge embeds: still render correctly
- Rate limit only affects >10 req/min from the same IP — normal badge embed usage unaffected
- `console.error` for actual GitHub API errors: preserved

Fixes GHSA-6c5j-4w43-2v8f vulnerability #4 (Medium).

### DIFF
--- a/src/app/api/badge/commits/route.ts
+++ b/src/app/api/badge/commits/route.ts
@@ -1,89 +1,85 @@
 import { NextRequest, NextResponse } from "next/server";
 import { generateBadgeSVG } from "../badge-utils";
+import { cacheGet, cacheSet } from "@/lib/metrics-cache";
 
 export const dynamic = "force-dynamic";
 
 const GITHUB_API = "https://api.github.com";
+const BADGE_CACHE_TTL = 3600;
 
-async function fetchGitHubWithToken(
-  url: string,
-  token?: string
-): Promise<Response> {
-  const headers: Record<string, string> = {
-    Accept: "application/vnd.github+json",
-  };
-
-  if (token) {
-    headers.Authorization = `Bearer ${token}`;
-  }
-
-  return fetch(url, { headers, cache: "no-store" });
-}
+// GitHub username: alphanumeric and hyphens, no leading/trailing hyphens,
+// no consecutive hyphens, 1–39 characters.
+const GITHUB_USERNAME_RE = /^[a-zA-Z0-9]([a-zA-Z0-9-]{0,37}[a-zA-Z0-9])?$/;
 
 async function fetchCommitsThisMonth(
   username: string,
   token?: string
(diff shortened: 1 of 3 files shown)


advikdivekar:

PR #645, opened 2026-05-22
test : added unit tests for isTruthyCacheBypass

### LINKED ISSUE #621: test : add unit tests for isTruthyCacheBypass function
Add unit tests for the isTruthyCacheBypass function in src/lib/metrics-cache.ts.

Changes that Need to be Made:
- Test with truthy values: '1', 'true', 'yes', 'on' (case insensitive)
- Test with falsy values: null, empty string, 'false', 'no'
- Test with whitespace handling

Impact that it would Provide:
- Ensures cache bypass logic works correctly
- Prevents cache bypass with invalid values

tmdeveloper007:
Closes #621.

Added unit tests for isTruthyCacheBypass.

Changes Made:
- Created test/metrics-cache.test.js with tests for isMetricsCacheBypassed

Impact it Made:
- Improves test coverage for utility functions

### DIFF
--- a/test/metrics-cache.test.js
+++ b/test/metrics-cache.test.js
@@ -1,0 +1,97 @@
+const assert = require("node:assert/strict");
+const test = require("node:test");
+
+function createMockNextRequest(url, headers = {}) {
+  const urlObj = new URL(url);
+  return {
+    nextUrl: {
+      searchParams: urlObj.searchParams,
+    },
+    headers: {
+      get: (name) => headers[name] || null,
+    },
+  };
+}
+
+test("isMetricsCacheBypassed returns true for refresh=1", async () => {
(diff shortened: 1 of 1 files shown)


Priyanshu-byte-coder (changes requested):
Two issues with this PR:

1. **Missing EOF newline** — test file must end with `\n`

2. **Tests don't import actual source code** — some tests copy-paste the implementation inside the test file (e.g., #662 redefines `getRateLimitRemainingFromResponse` locally). Tests must import and test the real function from `src/`. Reimplementing the function in the test only verifies the copy, not the actual code.

3. **No test runner configured** — the project has no `jest`, `vitest`, or `test` script in `package.json`. Please add vitest (or jest) config and a `test` script so these tests can actually run in CI. Reference: project uses TypeScript so `vitest` with `@vitest/ui` is the simplest fit.

Please fix all three and push again.

PR #761, opened 2026-05-23
fix : moved TTL validation guard before Redis check in cacheSet for universal coverage

### LINKED ISSUE #756: fix : move TTL validation in cacheSet to guard both memory and Redis paths
In src/lib/metrics-cache.ts, the cacheSet function contains a TTL validation guard that checks for invalid values (non-number, zero, negative, non-finite). However, this guard is only reached when no Redis client is available, meaning an invalid TTL such as 0, -1, NaN, or Infinity is silently forwarded to redis.set() when Redis is configured. The guard should be moved before the Redis/memory branch so it applies universally.

Changes that Need to be Made:
In src/lib/metrics-cache.ts, the current cacheSet function:

export async function cacheSet<T>(
  key: string,
  value: T,
  ttlSeconds: number
): Promise<void> {
  const redis = getRedisClient();
  
  if (redis) {
    try {
      await redis.set(key, value, { ex: ttlSeconds });  // invalid TTL reaches here
    } catch {
      // Cache failures must not break dashboard metrics.
    }
    return;
  }

if (typeof ttlSeconds !== 'number' || ttlSeconds <= 0 || !Number.isFinite(ttlSeconds)) {
    console.warn('Invalid TTL value:', ttlSeconds);
    return;
  }
  ...
}

Should be refactored to:

export async function cacheSet<T>(
  key: string,
  value: T,
  ttlSeconds: number
): Promise<void> {
  if (typeof ttlSeconds !== 'number' || ttlSeconds <= 0 || !Number.isFinite(ttlSeconds)) {
    console.warn('Invalid TTL value:', ttlSeconds);
    return;
  }

  const redis = getRedisClient();
  ...
}

Also fix the indentation of the TTL guard block (currently at column 0) to match the 2-space indentation style used throughout the file.

Impact that it would Provide:
Prevents silent forwarding of invalid TTL values to the Redis client, which could cause unexpected behavior or errors in the Upstash Redis backend. Ensures consistent validation regardless of which cache backend is active.

tmdeveloper007:
Closes #756.

Refactored the cacheSet function in src/lib/metrics-cache.ts to validate the TTL parameter before checking whether Redis is available. Previously, the TTL guard at line 121 was unreachable when Redis was configured because it appeared after the early return. Now the guard runs first, rejecting invalid TTL values (non-number, zero, negative, or non-finite) before any cache operation is attempted.

Changes Made:
Modified: src/lib/metrics-cache.ts

Key changes:
- Moved TTL validation guard from after the Redis/memory branch to before any cache client lookup
- Applied 2-space indentation to the guard block (previously at column 0)
- The guard now applies to both Redis and memory cache paths

Before:
```typescript
const redis = getRedisClient();
if (redis) {
  await redis.set(key, value, { ex: ttlSeconds });  // invalid TTL reaches here
  return;
}
if (typeof ttlSeconds !== "number" || ...) {  // only reached when no Redis
  console.warn("Invalid TTL:", ttlSeconds);
  return;
}
```

After:
```typescript
if (typeof ttlSeconds !== "number" || ttlSeconds <= 0 || !Number.isFinite(ttlSeconds)) {
  console.warn("Invalid TTL:", ttlSeconds);
  return;
}
const redis = getRedisClient();
if (redis) {
  await redis.set(key, value, { ex: ttlSeconds });
  return;
}
// memory cache path...
```

Impact it Made:
All existing tests pass (verified with npx vitest run). Prevents invalid TTL values from being silently forwarded to the Upstash Redis client. Ensures consistent validation regardless of which cache backend is active.

### DIFF
--- a/test/metrics-cache.test.ts
+++ b/test/metrics-cache.test.ts
@@ -1,5 +1,10 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 
+declare global {
+  // eslint-disable-next-line no-var
+  var metricsMemoryCache: Map<string, { value: unknown; expiresAt: number }> | undefined;
+}
+
 const mockRedisGet = vi.fn();
 const mockRedisSet = vi.fn();
 
@@ -141,7 +146,7 @@ describe('cacheGet', () => {
     const { cacheGet } = await import('../src/lib/metrics-cache');
     const result = await cacheGet('stale-key');
     expect(result).toBeNull();
-    const cached = globalThis.metricsMemoryCache;
+    const cached = globalThis.metricsMemoryCache!;
     expect(cached.has('stale-key')).toBe(false);
   });
 


Priyanshu-byte-coder (changes requested):

tmdeveloper007:

Priyanshu-byte-coder:
Merge conflict with current main. Please rebase against main to resolve before this can be merged.

Priyanshu-byte-coder:
This PR has merge conflicts with `main`. Please rebase on current main to resolve and re-request review. All changes look good otherwise.

PR #796, opened 2026-05-23
test : added unit tests for getRedisClient lazy initialization

### LINKED ISSUE #776: test : add unit tests for getRedisClient lazy initialization
getRedisClient lazily initializes the Redis client based on env vars but has no dedicated test. The null-client fallback behavior when UPSTASH_REDIS_REST_URL or UPSTASH_REDIS_REST_TOKEN are missing, and the module-level singleton caching, are untested.

Changes that Need to be Made:
Add tests to test/metrics-cache.test.ts covering:

getRedisClient:
- Returns null when UPSTASH_REDIS_REST_URL is not set
- Returns null when UPSTASH_REDIS_REST_TOKEN is not set
- Returns Redis client instance when both env vars are set
- Returns the same singleton instance on repeated calls (cache behavior)
- Calls constructor with correct url and token from env vars

Mock process.env for each test case. Use module-level mocking to reset redisClient between tests.

Impact that it would Provide:
Validates that the Redis client is correctly initialized only when credentials are available, and that the singleton pattern works. Prevents bugs where the client is re-created on every call.

tmdeveloper007:
Closes #776.

Added test/getRedisClient.test.ts with 4 vitest tests covering the getRedisClient lazy initialization from src/lib/metrics-cache.ts.

Changes Made:
New file: test/getRedisClient.test.ts

Test coverage:
- Returns null when UPSTASH_REDIS_REST_URL is not set
- Returns null when UPSTASH_REDIS_REST_TOKEN is not set
- Returns Redis instance when both env vars are set
- Returns same singleton instance on repeated calls (cache behavior)

Impact it Made:
All 4 tests pass. Redis client initialization validated.

### DIFF
--- a/src/lib/metrics-cache.ts
+++ b/src/lib/metrics-cache.ts
@@ -42,7 +42,7 @@ function isTruthyCacheBypass(value: string | null): boolean {
   return ["1", "true", "yes", "on"].includes(value.trim().toLowerCase());
 }
 
-function getRedisClient(): Redis | null {
+export function getRedisClient(): Redis | null {
   if (redisClient !== undefined) {
     return redisClient;
   }
--- a/test/getRedisClient.test.ts
+++ b/test/getRedisClient.test.ts
@@ -1,0 +1,50 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+vi.mock('@upstash/redis', () => {
+  const MockRedis = vi.fn(function() { return {}; });
+  return { Redis: MockRedis };
+});
+
+describe('getRedisClient lazy initialization', () => {
+  beforeEach(() => {
+    vi.resetModules();
+  });
+
+  it('returns null when UPSTASH_REDIS_REST_URL is not set', async () => {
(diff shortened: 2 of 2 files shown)


Priyanshu-byte-coder (changes requested):
Import pattern is correct — function imported from source. Three infrastructure issues to fix before merge:

1. **Missing `"test": "vitest run"`** in `package.json` scripts — tests can't run in CI without this.
2. **`vitest.config.ts` missing `resolve.alias`** — `@/lib/*` path aliases fail without:
```ts
import path from 'path'
export default defineConfig({
  resolve: { alias: { '@': path.resolve(__dirname, 'src') } },
  test: { globals: true }
})
```
3. **Missing EOF newline** on test file.

tmdeveloper007:

Priyanshu-byte-coder:
Merge conflict: `vitest.config.ts` already exists in main (added by PR #801). Please rebase against main to resolve before this can be merged.

PR #832, opened 2026-05-23
fix : use upsert with onConflict to handle race condition in streak freeze

tmdeveloper007:
fix : use upsert with onConflict to handle race condition in streak freeze

### DIFF
--- a/src/app/api/streak/freeze/route.ts
+++ b/src/app/api/streak/freeze/route.ts
@@ -70,23 +70,26 @@ export async function POST() {
 
   const today = todayStr();
 
+  const { data: existing } = await supabaseAdmin
+    .from("streak_freezes")
+    .select("id")
+    .eq("user_id", user.id)
+    .eq("freeze_date", today)
+    .maybeSingle();
+
   const { data: freeze, error } = await supabaseAdmin
     .from("streak_freezes")
-    .insert({ user_id: user.id, freeze_date: today })
+    .upsert({ user_id: user.id, freeze_date: today }, { onConflict: "user_id,freeze_date" })
     .select()
     .single();
 
   if (error) {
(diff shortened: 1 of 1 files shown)


Priyanshu-byte-coder (changes requested):
The upsert correctly eliminates the race condition, but removes the 409 response that clients relied on to detect 'already have an unused freeze'. Now both new and duplicate freezes return 200 — callers can no longer distinguish them. Add a flag to the response (e.g. `{ ...record, already_existed: boolean }`) or restore the 409 for the duplicate case.

Priyanshu-byte-coder:
The upsert approach is correct for handling race conditions, but removing the 409 response removes user feedback.

**Issue:** The previous code returned a helpful `409 Conflict` with message _"You already have an unused streak freeze."_ when a duplicate was attempted. After your change, both a race-condition duplicate **and** a genuine DB error both return `500 Failed to apply freeze.`

Please restore the 409 path. With upsert, you can check if the returned row's `created_at` matches the original insert (or simply keep a flag), but the simplest fix is: if upsert succeeds and returns data, treat it as success — but if no error occurs and data is returned, check whether it was an insert or an existing row by comparing with today's date against any existing freeze for the user. Alternatively, just restore the 409 for upsert no-ops:
```ts
// After upsert succeeds, check if freeze already existed
if (!error && freeze) {
  // upsert always succeeds — but if freeze_date already existed, it's a no-op update
  // to distinguish: query streak_freezes for this user before upsert, or use insert + catch 23505
}
```
The simplest approach is to keep `insert` and return 409 on error code `23505` (which your original code did correctly). The race condition upsert is fine, but users should still see the "already frozen" message.

tmdeveloper007:

PR #839, opened 2026-05-23
fix: validate githubId parameter in user github-accounts DELETE handler

tmdeveloper007:
fix: validate githubId parameter in user github-accounts DELETE handler


Priyanshu-byte-coder (changes requested):

PR #841, opened 2026-05-23
fix: prevent hydration mismatch in useHeatmapTheme by suppressing mismatch on theme element

tmdeveloper007:
fix: prevent hydration mismatch in useHeatmapTheme by suppressing mismatch on theme element


Priyanshu-byte-coder (changes requested):
Missing EOF newline on `e2e/landing.spec.js`. Also: the returned `suppressHydrationMismatch` flag is not used by any call site — either apply `suppressHydrationWarning` in the component that consumes this hook, or remove the flag from the return value.

Priyanshu-byte-coder:
One issue to fix:

**Missing EOF newline in `e2e/landing.spec.js`**
The diff shows `-});\n+});\ No newline at end of file` — your change removes the trailing newline from the file. All files must end with a newline. Please restore it.

The hydration mismatch fix itself (`mounted` state, returning `"default"` until mounted) is correct and the approach is good.

tmdeveloper007:

PR #842, opened 2026-05-23
fix: user settings PATCH treats null body values as field omission

tmdeveloper007:
fix: user settings PATCH treats null body values as field omission

### DIFF
--- a/src/app/api/user/settings/route.ts
+++ b/src/app/api/user/settings/route.ts
@@ -157,21 +157,25 @@ export async function PATCH(req: NextRequest) {
   }
 
   const { hasLeaderboardOptIn, hasPinnedRepos } = settingsResult;
-
   const updates: { is_public?: boolean; leaderboard_opt_in?: boolean; pinned_repos?: string[] } = {};
 
-  if (typeof is_public === "boolean") {
+  if (is_public !== undefined && is_public !== null && typeof is_public === "boolean") {
     updates.is_public = is_public;
   }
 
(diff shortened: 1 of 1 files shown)


Priyanshu-byte-coder (changes requested):
Two issues:

1. **Missing EOF newline** on `e2e/landing.spec.js`.

2. **Null allowed into boolean field**: Changing guards from `typeof is_public === 'boolean'` to `is_public !== undefined` allows `{ is_public: null }` through — `null` will be assigned to a typed boolean column. Change to `is_public !== undefined && is_public !== null` to properly handle field omission vs explicit null.

Priyanshu-byte-coder:
Three issues block this merge:

**1. `!== undefined` check is too permissive**
Changing `typeof is_public === "boolean"` to `is_public !== undefined` means `is_public: null` or `is_public: 1` would pass through and be written to the DB. The strict `typeof === "boolean"` check is intentional — please revert those changes.

Same problem with `pinned_repos`: replacing `Array.isArray()` with `!== undefined` means a string value like `pinned_repos: "foo"` would pass the `if (pinned_repos !== undefined)` check, and then `pinned_repos.length` would measure string length instead of array length.

**2. Merge conflict**
This PR conflicts with current main (`settings/route.ts` was updated by a recently merged PR). Please rebase on main.

**3. Missing EOF newline in `e2e/landing.spec.js`**
The diff shows the trailing newline removed. Please restore it.

tmdeveloper007:

PR #904, opened 2026-05-24
fix: prune stale leaderboard cache entries

### LINKED ISSUE #806: fix : add TTL-based eviction to leaderboard in-memory cache Maps
In src/app/api/leaderboard/route.ts, both leaderboardCache and ipRateLimits are never pruned. The ipRateLimits Map has no eviction, and while leaderboardCache has an expiresAt, expired entries are only removed on access (not proactively). Long-running server instances accumulate unbounded Map entries.

Changes that Need to be Made:
Add proactive cleanup for both Maps:

```typescript
// At start of each request:
const now = Date.now();
// Clean ipRateLimits
for (const [key, val] of ipRateLimits.entries()) {
  if (val.resetAt <= now) ipRateLimits.delete(key);
}
// Clean leaderboardCache
for (const [key, val] of leaderboardCache.entries()) {
  if (val.expiresAt <= now) leaderboardCache.delete(key);
}
```

Impact that it would Provide:
Prevents unbounded memory growth in long-running server instances. Ensures stale cache entries are cleaned up proactively.

saurabhhhcodes:
## Summary
- proactively prune expired IP rate-limit buckets on every leaderboard request
- move leaderboard cache eviction into a small reusable helper so stale payloads are cleared consistently
- add a focused regression test for rate-limit and cache cleanup behavior

Closes #806

## Validation
- `node --experimental-strip-types --test test/leaderboard-cache.test.mjs`
- `./node_modules/.bin/tsc --noEmit`
- `git diff --check`

### DIFF
--- a/src/app/api/leaderboard/route.ts
+++ b/src/app/api/leaderboard/route.ts
@@ -1,6 +1,12 @@
 import { NextRequest, NextResponse } from "next/server";
 import { supabaseAdmin } from "@/lib/supabase";
 import { dateDiffDays, toDateStr } from "@/lib/dateUtils";
+import {
+  pruneExpiredLeaderboardCache,
+  pruneExpiredRateLimits,
+  type LeaderboardCacheEntry,
+  type RateLimitEntry,
+} from "@/lib/leaderboard-cache";
 
 export const dynamic = "force-dynamic";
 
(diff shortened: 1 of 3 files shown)


saurabhhhcodes:
Pushed a follow-up fix for the failing typecheck: the activity route now narrows session account fields into local non-optional values before the combined-account branches use them. Local `tsc --noEmit` passes on the updated branch.

PR #913, opened 2026-05-24
test: add unit tests for metricsCacheKey edge cases and cacheGet miss behavior

### LINKED ISSUE #825: fix : add unit tests for cacheGet cache miss behavior
The cacheGet function in metrics-cache.ts has no unit tests for the cache miss path. Specifically, when Redis is available but a key is not found, and when memory cache has an expired entry.

Changes that Need to be Made:
Add tests to test/metrics-cache.test.ts covering:

- Returns null when key not found in Redis
- Returns null when key not found in memory cache
- Deletes expired entry from memory cache (side effect)
- Falls back to memory cache when Redis returns null

Impact that it would Provide:
Validates cache miss handling. Ensures expired entries are cleaned up.

### LINKED ISSUE #826: test : add unit tests for metricsCacheKey param serialization
The metricsCacheKey function in metrics-cache.ts has existing tests for boolean params, but there are edge cases not covered: null values are filtered (correct), undefined values are filtered (correct), numeric zero is included in key (correct), empty string is included (might be a bug).

Changes that Need to be Made:
Add additional test cases to test/metrics-cache.test.ts:

- null param values are correctly filtered out (not included in key)
- undefined param values are correctly filtered out
- numeric zero (0) is included as '0' in key
- empty string is included as '' in key

Impact that it would Provide:
Validates edge case param serialization behavior. Catches any future regression in null/undefined filtering.

Akshita-2307:
Adds unit tests for two areas missing coverage in `metrics-cache.ts`:

**metricsCacheKey edge cases** (fixes #826):
- null param values are filtered out of the cache key
- undefined param values are filtered out
- numeric zero is included as `"0"` in the key
- empty string is included in the key

**cacheGet miss/fallback behavior** (fixes #825):
- Returns null when Redis is unavailable and memory cache is empty
- Returns null when Redis is available but key not found (catch path)
- Deletes expired memory cache entries and returns null
- Falls back to memory cache when Redis is unavailable with valid cached data

All 8 tests pass using the same `node:test` + TypeScript transpilation pattern as the existing `test/crypto.test.js`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,4 +34,7 @@ Thumbs.db
 # Claude Code local settings
 .claude/
 
+# Test generated output
+test/__generated__/
+
 desktop.ini
--- a/test/metrics-cache.test.js
+++ b/test/metrics-cache.test.js
@@ -1,0 +1,106 @@
+const assert = require("node:assert/strict");
+const fs = require("node:fs");
+const path = require("node:path");
+const test = require("node:test");
+const ts = require("typescript");
+
+function loadMetricsCacheModule() {
+  const sourcePath = path.join(__dirname, "..", "src", "lib", "metrics-cache.ts");
+  const outDir = path.join(__dirname, "__generated__");
+  fs.mkdirSync(outDir, { recursive: true });
+  const outPath = path.join(outDir, "metrics-cache.cjs");
+  const source = fs.readFileSync(sourcePath, "utf8");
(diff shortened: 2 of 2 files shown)


Akshita-2307:
Superseded by #921 which uses vitest (matching the repo's new test framework) and covers more edge cases including booleans and TTL validation.

PR #1207, opened 2026-05-26
fix(local-coding): enforce per-user cap of 10 API keys

### LINKED ISSUE #981: [BUG] No per-user cap on API key creation in /api/local-coding/keys — storage exhaustion and auth performance degradation
## Describe the Bug

The `POST /api/local-coding/keys` handler inserts a new API key row for the authenticated user with no limit on how many keys a single user can create. There is no count check anywhere in the handler before the `insert` call.

**Relevant code in `src/app/api/local-coding/keys/route.ts`:**
```ts
// No count guard — insert runs unconditionally
const { data: keyRecord, error } = await supabaseAdmin
  .from('local_coding_api_keys')
  .insert({
    user_id: user.id,
    api_key: apiKeyHash,
    name
  })
```

This is inconsistent with the `local-coding/sync` route which enforces both `MAX_SESSIONS_PER_REQUEST = 100` and `MAX_SESSIONS_PER_USER = 365` before any insert.

**Impact:**

1. **Storage exhaustion:** Any authenticated user can script unlimited `POST /api/local-coding/keys` calls, inserting thousands of rows in seconds.
2. **Auth performance degradation for all users:** The `authenticateApiKey` function used on every `/api/local-coding/sync` request does a full-table lookup via `.eq('api_key_hash', hash)`. As a single user's key count grows into the thousands, that table scan slows down for every user hitting the sync endpoint.

## Steps to Reproduce

```bash
# Authenticated session cookie required
for i in $(seq 1 500); do
  curl -X POST https://<your-devtrack>/api/local-coding/keys \
    -H 'Content-Type: application/json' \
    -H 'Cookie: <session>' \
    -d "{\"name\": \"key-$i\"}"
done
```

After this loop, the `local_coding_api_keys` table will have 500 rows for a single user and the sync endpoint's authentication will visibly slow down.

## Expected Behavior

The endpoint should enforce a reasonable per-user cap (e.g. 10 keys) and return HTTP 400 with a message like `"Maximum of 10 API keys allowed per user"` when the limit is reached.

## Screenshots / Logs

No immediate error — the bug manifests as silent table bloat and gradual performance degradation.

## Environment

- OS: Any
- Node version: 18+
- Browser: Any

## Additional Context

The fix mirrors the pattern already in `src/app/api/local-coding/sync/route.ts`:
1. Add `const MAX_KEYS_PER_USER = 10` constant
2. Before the insert: query `SELECT count(*) FROM local_coding_api_keys WHERE user_id = user.id`
3. Return 400 if count meets the cap

I am happy to work on this if it can be assigned to me.

anshul23102:
## Summary

Fixes #981

The POST `/api/local-coding/keys` handler had no guard on how many API keys a single user could create. This allowed unbounded key creation, causing:

- **Storage bloat** in the `local_coding_api_keys` table
- **Auth performance degradation**: `authenticateApiKey` (in the sync route) does a full scan of keys per user on every sync request, so a user with hundreds of keys slows down all their sync calls

## Changes

- Added `MAX_KEYS_PER_USER = 10` constant (mirrors the `MAX_SESSIONS_PER_USER = 365` pattern in `src/app/api/local-coding/sync/route.ts`)
- Before inserting a new key, query the exact count of existing keys for the user
- Return HTTP 400 with a clear message if the limit is already reached

## Behavior

| Scenario | Before | After |
|---|---|---|
| User has < 10 keys | Key created | Key created (no change) |
| User has 10 keys | Key created (unbounded) | 400 - "API key limit reached. Maximum 10 keys per user." |

## Testing

- Create keys up to the limit: all succeed
- Create one more: returns 400 with the limit message
- Delete a key then create again: succeeds (count drops below limit)


Priyanshu-byte-coder:
CI is failing on this PR. Please fix the failing checks before this can be merged.

anshul23102:
Hi @Priyanshu-byte-coder, just wanted to follow up on this PR. All CI checks are now passing (Build, Lint, Type check, Playwright smoke tests, Dependency audit, and GitGuardian). Could you take a look when you get a chance? Happy to address any feedback. Thank you!

PR #1250, opened 2026-05-27
fix: replace any type with SupabaseClient | null for type-safe null h…

### LINKED ISSUE #1249: [Bug] supabaseAdmin typed as any allowing silent null crashes across all API routes
## Bug Description
`supabaseAdmin` in `src/lib/supabase.ts` is typed as `any` despite 
being explicitly set to `null` when environment variables are missing. 
This means TypeScript provides zero protection — any API route calling 
`supabaseAdmin.from(...)` without a null check will throw an unhandled 
`TypeError: Cannot read properties of null` at runtime, crashing the 
entire route handler.

## Steps to Reproduce
1. Remove or leave empty `NEXT_PUBLIC_SUPABASE_URL` in `.env.local`
2. Hit any API route that uses `supabaseAdmin` directly
3. Observe: unhandled `TypeError` crashes the route instead of a 
   clean error response

## Expected Behavior
TypeScript should catch unguarded `supabaseAdmin` calls at compile 
time. The type should reflect that it can be `null` so developers 
are forced to handle that case.

## Root Cause
In `src/lib/supabase.ts`:

```ts
export const supabaseAdmin: any = ...
```

Typing it as `any` disables all TypeScript checks. The correct type 
is `SupabaseClient | null` which forces null checks at every call site.

## Fix
Import `SupabaseClient` and use it as the type:

```ts
import { createClient, SupabaseClient } from "@supabase/supabase-js";

export const supabaseAdmin: SupabaseClient | null =
  supabaseUrl && serviceRoleKey && !supabaseUrl.includes("placeholder")
    ? createClient(supabaseUrl, serviceRoleKey)
    : null;
```

## File Affected
`src/lib/supabase.ts`

## Impact
Every API route in the app that uses `supabaseAdmin` is affected.

akhilmodi29:
## What does this PR fix?
Closes #1249 
`supabaseAdmin` was typed as `any` despite being explicitly set to 
`null` when environment variables are missing. This disabled all 
TypeScript protection — unguarded calls to `supabaseAdmin.from(...)` 
would throw unhandled `TypeError` crashes at runtime across every 
API route in the app.

## Root Cause
```ts
export const supabaseAdmin: any = ...
```
Typing as `any` silences TypeScript completely. The value can be 
`null` at runtime but nothing warns you at compile time.

## Changes Made
- Imported `SupabaseClient` type from `@supabase/supabase-js`
- Replaced `any` with `SupabaseClient | null`
- TypeScript now catches unguarded `supabaseAdmin` calls at compile time

## Files Changed
- `src/lib/supabase.ts`

## Testing Done
1. Run `npm run type-check`
2. ✅ TypeScript correctly flags unguarded supabaseAdmin calls
3. ✅ Existing null-guarded call sites pass type check
4. ✅ No runtime behaviour changed


akhilmodi29:
Hi @Priyanshu-byte-coder, please review this PR and add the required 
labels including gssoc:approved. This fixes supabaseAdmin being typed 
as `any` despite being explicitly nullable — replacing it with 
`SupabaseClient | null` so TypeScript catches unguarded null calls 
at compile time across all API routes.


Labels suggested:
- level:intermediate
- type:bug
- quality:clean
- gssoc:approved

Priyanshu-byte-coder:
Hi @akhilmodi29! Thank you for this improvement. We've merged a more comprehensive version (PR #1263 by @mittalsonal) that goes further — instead of typing the null as `SupabaseClient | null`, it replaces null with a proxy object that throws a descriptive error, avoiding the need for null checks at call sites entirely. Your approach was valid; the other PR just covers more ground. Thanks for contributing!

PR #1262, opened 2026-05-27
fix: restore goal form after failed refresh in create flow

### LINKED ISSUE #1243: [Bug] Goal form submit button stays disabled after a failed create attempt
## Bug Description
When creating a goal fails (network error, server error, or if the 
post-create sync throws), the "Add goal" submit button becomes permanently 
disabled for the rest of the session. The user cannot retry without 
refreshing the page.

## Steps to Reproduce
1. Go to dashboard → Goal Tracker widget
2. Fill in the goal creation form
3. Disconnect internet → click "Add goal"
4. Observe: button stays stuck in "Creating..." state permanently

## Expected Behavior
Button should return to its normal enabled state after any failure 
so the user can retry.

## Root Cause
In `handleCreate()` inside `GoalTracker.tsx`, `setCreating(false)` 
is only called at the end of the happy path. If `handleSync()` throws 
after a successful POST, `setCreating(false)` never runs.

## Fix
Move form reset logic inside `try` block and use `finally` to 
guarantee `setCreating(false)` always runs.

## File Affected
`src/components/GoalTracker.tsx` — `handleCreate()` function

mittalsonal:
## Summary
- move goal creation and follow-up refresh into a small reusable helper
- always release the create button state even when the post-create sync/reload step fails
- add focused tests covering create failure, refresh failure, and non-auto-synced goal reloads

## Testing
- git diff --check
- bundled Node smoke test for submitGoalWithRefresh covering failed create, failed refresh, and successful reload paths

Closes #1243

### DIFF
--- a/src/components/GoalTracker.tsx
+++ b/src/components/GoalTracker.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import { useCallback, useEffect, useState, useRef } from "react";
+import { submitGoalWithRefresh } from "@/lib/goal-tracker";
 
 type Recurrence = "none" | "weekly" | "monthly";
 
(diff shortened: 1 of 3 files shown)


mittalsonal:
Verification completed here:
- git diff --check
- bundled Node smoke test covering failed create, failed refresh, and successful non-auto-synced reload paths for submitGoalWithRefresh

I could not run the full repo itest/	sc commands in this workspace because the checkout had no local dependencies installed and the machine hit a disk-space limit during 
pm ci. The patch itself is scoped to the goal-create flow and includes a focused regression test file for the maintainers' normal CI run.

mittalsonal:
Requesting GSSoC review labels when convenient: gssoc:approved, one difficulty label, and the relevant type label for this fix. Thank you!

mittalsonal:
Rebased this branch onto the current main and force-pushed the updated head so the stale-branch blocker should be cleared. No code scope changed beyond replaying the same fix on top of latest upstream.

Priyanshu-byte-coder:
Merged manually as a squash commit to main. Thank you for your contribution! 🎉

PR #1263, opened 2026-05-27
fix: guard unavailable supabase admin client

### LINKED ISSUE #1249: [Bug] supabaseAdmin typed as any allowing silent null crashes across all API routes

mittalsonal:
## Summary
- replace the nullable ny admin client with a typed guarded client path
- throw a clear configuration error when the admin client is unavailable instead of allowing a 
ull.from(...) crash
- add a focused regression test covering the unavailable-client guard behavior

## Testing
- git diff --check
- static Node checks confirming the new guard constant/factory and the new regression test coverage shape

Closes #1249

### DIFF
--- a/src/lib/supabase.ts
+++ b/src/lib/supabase.ts
@@ -3,12 +3,25 @@ import { createClient } from "@supabase/supabase-js";
 const supabaseUrl = process.env.NEXT_PUBLIC_SUPABASE_URL;
 const serviceRoleKey = process.env.SUPABASE_SERVICE_ROLE_KEY;
 
-// Do not throw here — build-time rendering can touch this module before
+export const SUPABASE_ADMIN_UNAVAILABLE_MESSAGE =
+  "Supabase admin client is unavailable. Check NEXT_PUBLIC_SUPABASE_URL and SUPABASE_SERVICE_ROLE_KEY.";
+
+type SupabaseAdminClient = ReturnType<typeof createClient>;
+
+function createUnavailableSupabaseAdmin(): SupabaseAdminClient {
+  return {
+    from() {
+      throw new Error(SUPABASE_ADMIN_UNAVAILABLE_MESSAGE);
+    },
+  } as unknown as SupabaseAdminClient;
+}
+
+// Do not throw here - build-time rendering can touch this module before
 // runtime environment variables are present. Guard call sites instead.
-export const supabaseAdmin: any =
+export const supabaseAdmin: SupabaseAdminClient =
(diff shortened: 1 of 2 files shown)


mittalsonal:
Verification completed here:
- git diff --check
- static Node checks confirming the new unavailable-client guard constant/factory and the new regression test coverage shape

I did not run the full repo itest/	sc commands in this workspace because dependencies are not installed here and the machine is extremely low on disk space, so I kept the patch narrowly scoped to the Supabase guard path plus one targeted test file.

mittalsonal:
Requesting GSSoC review labels when convenient: gssoc:approved, one difficulty label, and the relevant type label for this guard fix. Thank you!

mittalsonal:
Rebased this branch onto the current main and force-pushed the updated head so the stale-branch blocker should be cleared. No code scope changed beyond replaying the same Supabase guard fix on top of latest upstream.

Priyanshu-byte-coder:
Merged manually as a squash commit to main. Thank you for your contribution! 🎉

PR #1403, opened 2026-05-28
fix(session): redirect to sign-in when session expires silently

### LINKED ISSUE #930: [BUG] Session expires silently — dashboard shows stale data with no redirect
## Describe the Bug

When the NextAuth session expires mid-session, API calls return 401 but the dashboard silently shows the last cached/empty data with no prompt to re-login. Users have no indication their data is stale.

## Steps to Reproduce

1. Log in and open the dashboard
2. Wait for session to expire (or manually delete the session cookie)
3. Click Refresh on any widget (e.g. activity feed)
4. No redirect to login — widget shows old data or empty state

## Expected Behavior

Any 401 response from API routes should trigger redirect to sign-in page with a toast: "Session expired. Please sign in again."

## Screenshots / Logs

Network tab shows 401 on `/api/metrics/*` with no visible UI change.

## Environment

- OS: Any
- Node version: 18+
- Browser: Any

## Additional Context

Could be handled globally in a fetch wrapper that intercepts 401s, or via NextAuth `useSession` with `required: true` in the dashboard layout.

mallya-m:
## What does this PR do?
Fixes silent 401 failures on the dashboard. When any API call returns 401 (session expired), the user now sees a toast "Session expired. Please sign in again." and is redirected to the sign-in page.

## Related issue
Closes #930

## Changes made
- Added `src/app/dashboard/layout.tsx` scoped to /dashboard routes
- Uses `useSession({ required: true })` as primary session guard
- Global fetch interceptor detects 401 responses from any API call
- Shows sonner toast (already in project) before redirecting
- Restores original fetch on unmount — no memory leaks
- No changes to existing components

## How to test
1. Log in and open the dashboard
2. Delete the session cookie in DevTools
3. Click Refresh on any widget
4. Toast appears: "Session expired. Please sign in again."
5. Automatically redirected to /auth/signin

### DIFF
--- a/src/app/dashboard/layout.tsx
+++ b/src/app/dashboard/layout.tsx
@@ -11,12 +11,10 @@ export default function DashboardLayout({ children }: { children: ReactNode }) {
   const router = useRouter();
 
   useEffect(() => {
-    // intercept all fetch calls — redirect on 401
     const originalFetch = window.fetch;
     window.fetch = async (...args) => {
       const response = await originalFetch(...args);
       if (response.status === 401) {
-        // clone so body can still be read by caller
         const cloned = response.clone();
         toast.error("Session expired. Please sign in again.");
         await signOut({ redirect: false });
--- a/src/app/dashboard/layout.tsx
+++ b/src/app/dashboard/layout.tsx
@@ -1,0 +1,36 @@
+"use client";
+
+import { useSession, signOut } from "next-auth/react";
+import { useEffect } from "react";
+import { useRouter } from "next/navigation";
+import { toast } from "sonner";
+import type { ReactNode } from "react";
+
+export default function DashboardLayout({ children }: { children: ReactNode }) {
+  const { status } = useSession({ required: true });
+  const router = useRouter();
+
+  useEffect(() => {
+    // intercept all fetch calls — redirect on 401
+    const originalFetch = window.fetch;
+    window.fetch = async (...args) => {
+      const response = await originalFetch(...args);
+      if (response.status === 401) {
+        // clone so body can still be read by caller
+        const cloned = response.clone();
+        toast.error("Session expired. Please sign in again.");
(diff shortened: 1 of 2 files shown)


mallya-m:
@Priyanshu-byte-coder could you please add `gssoc:approved`, `level:intermediate`, and `type:bug` labels? 
Ready to merge .Thank you!

Priyanshu-byte-coder (changes requested):
Thanks for working on session expiry handling! The concept is right but the implementation has issues:

**`window.fetch` monkey-patching is risky:**
```ts
window.fetch = async (...args) => {
  const response = await originalFetch(...args);
  if (response.status === 401) { // intercepts ALL fetch calls
```
This intercepts every fetch in the dashboard including third-party API calls (GitHub, etc.). A 401 from GitHub's API would incorrectly sign the user out of DevTrack.

**Better approach:**
1. Use NextAuth's middleware (already in `src/middleware.ts`) to redirect unauthenticated users
2. For specific session-expiry UX, use `useSession({ required: true, onUnauthenticated() { router.push('/auth/signin') } }` without the fetch intercept

Please revise to avoid monkey-patching the global fetch.

Priyanshu-byte-coder (changes requested):
The session-expiry redirect logic is useful but the implementation monkey-patches `window.fetch` globally in a React component. This is a risky pattern that can cause flaky behavior and interfere with other fetches on the page.

**Preferred approach:** Use a custom `fetch` wrapper utility or an Axios interceptor in a dedicated module (e.g. `src/lib/auth-fetch.ts`), then import it where needed. Or use `next-auth`'s `useSession({ required: false })` combined with an `error === 'TokenRevoked'` check (already supported by the JWT callback) to trigger redirect without patching globals.

Please refactor to avoid patching `window.fetch` and rebase on `main`.

PR #1496, opened 2026-05-29
test: add validation unit tests for user settings API endpoints 

### LINKED ISSUE #818: fix : user settings PATCH should treat null body values as field omission
In src/app/api/user/settings/route.ts PATCH handler, if body contains { is_public: null }, the update will set is_public to null rather than ignoring the null. The current typeof check only verifies the value is not an object, so null passes through.

Changes that Need to be Made:
Use undefined check instead of typeof check to distinguish between field omission and explicit null:

```typescript
const { is_public, ...rest } = body as Record<string, unknown>;
const updates: Record<string, unknown> = {};
if (is_public !== undefined) updates.is_public = is_public;
```

Impact that it would Provide:
Prevents accidental null assignment when client sends { is_public: null }. Null fields are explicitly excluded from updates.

bhavyanjain3004:
### Description
This PR adds comprehensive unit test coverage for the user settings API (`GET` and `PATCH` endpoints in `src/app/api/user/settings/route.ts`). Specifically, it verifies that `null` values for settings (such as `is_public`, `leaderboard_opt_in`, and `pinned_repos`) are ignored/treated as field omissions in the `PATCH` body, preventing database updates to those fields.

Closes #818

### Changes
- Added [user-settings-api.test.ts]covering:
  - Authentication checks and error handling for GET/PATCH.
  - Verification that maximum limits (3 pinned repos) are enforced.
  - Verification that passing `null` fields in the PATCH request body is ignored and doesn't trigger database updates.
  - Verification that valid settings updates are properly mapped and saved.


Priyanshu-byte-coder:
Merged manually into main via squash commit with full co-authorship credit. Thank you @bhavyanjain3004 for the contribution! 🎉

PR #1526, opened 2026-05-29
fix(local-coding): align API key hash storage

### LINKED ISSUE #1524: [BUG] Local coding API keys are stored in api_key but sync checks api_key_hash
## Line
`src/app/api/local-coding/keys/route.ts:70-75`
`src/app/api/local-coding/sync/route.ts:19-23`
`supabase/migrations/20260522000000_add_api_key_hash_column.sql:1`

## Problem
I ran into this while following the local coding API key flow for GSSoC 2026. New keys are generated and hashed in `POST /api/local-coding/keys`, but the hash is inserted into the `api_key` column:

```ts
.insert({
  user_id: user.id,
  api_key: apiKeyHash,
  name,
})
```

The sync route authenticates the same key by looking up `api_key_hash` instead:

```ts
.eq("api_key_hash", keyHash)
```

Since the migration added a separate nullable `api_key_hash` column, newly created keys can end up with `api_key_hash = null`. The key is returned to the user once, but `/api/local-coding/sync` will not find it later.

## Impact
This breaks the main local coding sync flow after a user creates a fresh API key. From the user's side, key creation looks successful, but the CLI/client sync request gets `401 Invalid API key`.

This also makes debugging confusing because the generated key itself is not wrong; it was just written into a different column than the auth path reads from.

## Reproduction
1. Sign in and create a local coding API key from the app, or call `POST /api/local-coding/keys` with a valid session.
2. Copy the returned key.
3. Send a sync request with `Authorization: Bearer <returned-key>` to `/api/local-coding/sync`.
4. The sync route hashes that key and checks `local_coding_api_keys.api_key_hash`, but the create route populated `api_key`.
5. The request returns `401 Invalid API key` even though the key was just created.

Expected behavior: the generated key should authenticate against the sync endpoint.

Current behavior: the key is stored in `api_key`, while sync checks `api_key_hash`, so authentication can fail for new keys.

## Suggested fix
Write the generated hash to `api_key_hash` when creating keys, and keep the sync route using `api_key_hash`. If `api_key` is still needed for backward compatibility, either backfill/copy existing hashes or make the auth lookup tolerate old rows during migration.

A focused regression test should cover creating a key and authenticating the returned plain key through the sync route lookup path.

## Labels
bug, needs-triage, gssoc26, level:intermediate, type:bug

eshaanag:
## Summary
- Store newly generated local coding API key hashes in `api_key_hash` as well as the existing `api_key` column.
- Let sync authentication match either `api_key_hash` or the legacy `api_key` column so existing rows keep working.
- Add regression coverage for key creation storage and sync authentication lookup behavior.

## Why
This fixes #1524 as part of GSSoC 2026. The key creation route was writing the generated hash into `api_key`, while the sync route authenticated against `api_key_hash`. That made a freshly generated key look valid in the UI but fail later with `401 Invalid API key` during sync.

## Testing
- `npm run test -- test/local-coding-keys.test.ts test/local-coding-sync.test.ts`
- `npm run lint`
- `npm run type-check`
- `node scripts/check-deps.js`
- `npm ci --dry-run`
- `npx -p node@20 -c 'node -v && npm run build'`

Closes #1524

### DIFF
--- a/src/app/api/local-coding/keys/route.ts
+++ b/src/app/api/local-coding/keys/route.ts
@@ -72,6 +72,7 @@ export async function POST(req: NextRequest) {
     .insert({
       user_id: user.id,
       api_key: apiKeyHash,
+      api_key_hash: apiKeyHash,
       name,
     })
     .select("id, name, last_used_at, created_at")
--- a/src/app/api/local-coding/sync/route.ts
+++ b/src/app/api/local-coding/sync/route.ts
@@ -15,11 +15,12 @@ function hashApiKey(key: string): string {
 
 async function authenticateApiKey(apiKey: string): Promise<string | null> {
   const keyHash = hashApiKey(apiKey);
+  const keyHashFilter = `api_key_hash.eq.${keyHash},api_key.eq.${keyHash}`;
 
   const { data: keyRecord } = await supabaseAdmin
     .from("local_coding_api_keys")
     .select("user_id")
-    .eq("api_key_hash", keyHash)
+    .or(keyHashFilter)
     .single();
 
   if (!keyRecord) {
@@ -29,7 +30,7 @@ async function authenticateApiKey(apiKey: string): Promise<string | null> {
   await supabaseAdmin
     .from("local_coding_api_keys")
     .update({ last_used_at: new Date().toISOString() })
-    .eq("api_key_hash", keyHash);
+    .or(keyHashFilter);
 
   return keyRecord.user_id;
 }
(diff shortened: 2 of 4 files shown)


eshaanag:
Quick CI note: the main CI gate is green here — type check, lint, build, dependency audit, and the CI summary all passed.\n\nThe only failing check is the Playwright smoke suite. The failures are in dashboard/widgets, notifications, settings, and theme specs, with NextAuth logging `Invalid Compact JWE` during the run. This PR only changes the local-coding API key storage/auth path plus focused unit tests, so the E2E failure does not look related to this patch.\n\nI tried to rerun the failed workflow once to rule out flakiness, but GitHub requires repository admin permissions for that action.

eshaanag:
Thanks for merging this. Could you please add the appropriate difficulty label and `gssoc:approved` if this PR is accepted for GSSoC 2026 scoring? The PR currently has `gssoc26`, `type:bug`, and `type:testing`, but no scoring approval/difficulty label yet.

PR #1531, opened 2026-05-29
fix(local-coding): count only new sync dates

### LINKED ISSUE #1530: [BUG] Local coding sync session cap counts upserts as new rows
## Line
`src/app/api/local-coding/sync/route.ts:88-91`
`src/app/api/local-coding/sync/route.ts:111-114`
`supabase/migrations/20260521000000_add_local_coding_tables.sql:10`

## Problem
I noticed this while testing the local coding sync path for GSSoC 2026. The sync route counts all rows a user already has, then adds the full incoming `newSessions.length` before deciding whether the user is over `MAX_SESSIONS_PER_USER`.

That is not the same as the number of rows that will be inserted. The table has `unique(user_id, date)`, and the route later calls `batch_upsert_sessions`, so syncing an already-existing date updates the row instead of creating a new one.

Right now, a user with 360 existing local coding days can be blocked from syncing 10 days even if all 10 dates already exist and would only be updates.

## Impact
Legitimate re-syncs can fail with `Session limit reached` near the 365-day cap. That breaks normal retry/backfill behavior for local coding clients, especially if a client sends the last 7/30/90 days every time.

Expected behavior: only truly new dates should count against the 365-session cap.

Current behavior: every valid incoming session is counted as new, even when the RPC would upsert an existing `(user_id, date)` row.

## Reproduction
1. Have a user with 360 rows in `local_coding_sessions`.
2. Send `POST /api/local-coding/sync` with 10 sessions whose dates already exist for that user.
3. The route calculates `360 + 10 > 365` and returns `400 Session limit reached`.
4. None of the existing rows are updated, although the database upsert would not increase the row count.

## Suggested fix
After validating the incoming sessions, collect their unique dates and query which of those dates already exist for the user. Only add the count of missing dates to `existingCount` when enforcing `MAX_SESSIONS_PER_USER`.

A regression test should cover the near-limit case where all incoming dates already exist and should be allowed to sync.

## Labels
bug, needs-triage, gssoc26, level:intermediate, type:bug

eshaanag:
## Summary
- Count only truly new incoming local-coding dates against `MAX_SESSIONS_PER_USER`.
- Keep existing-date re-syncs allowed near the 365-session cap.
- Add regression coverage for the near-limit upsert case.

## Why
This fixes #1530 as part of GSSoC 2026. The sync route was adding the full incoming session count to the user's existing row count before enforcing the cap, but the database upserts by `(user_id, date)`. That meant normal re-syncs of existing dates could be rejected even though they would not create new rows.

## Testing
- `npm run test -- test/local-coding-sync.test.ts`
- `npm run lint`
- `npm run type-check`
- `node scripts/check-deps.js`
- `npm ci --dry-run`
- `npx -p node@20 -c 'node -v && npm run build'`

Note: the Node 20 build completed successfully. Next.js still prints the existing upstream `package-lock.json` parse warning from conflict markers, but this PR does not touch the lockfile.

Closes #1530

### DIFF
--- a/e2e/dashboard-widgets.spec.js
+++ b/e2e/dashboard-widgets.spec.js
@@ -224,7 +224,10 @@ test("contribution graph range buttons request a new range", async ({ page }) =>
 
   await page.goto("/dashboard", { waitUntil: "load" });
   await expect(page.getByRole("heading", { name: /dashboard/i })).toBeVisible({ timeout: 30000 });
-  await page.getByRole("button", { name: "Show 90-day range" }).click();
+  await page
+    .locator("#contribution-activity")
+    .getByRole("button", { name: "Show 90-day range" })
+    .click();
 
   await expect.poll(() => contributionRequests.some((url) => url.includes("days=90")), { timeout: 15000 }).toBe(true);
 });
--- a/playwright.config.mjs
+++ b/playwright.config.mjs
@@ -2,6 +2,8 @@ import { defineConfig, devices } from "@playwright/test";
 
 const PORT = Number(process.env.PORT ?? 3000);
 const baseURL = process.env.PLAYWRIGHT_BASE_URL ?? `http://127.0.0.1:${PORT}`;
+const prepareStandaloneCommand =
+  "node -e \"const fs=require('fs'); fs.cpSync('public','.next/standalone/public',{recursive:true,force:true}); fs.cpSync('.next/static','.next/standalone/.next/static',{recursive:true,force:true});\"";
 
 export default defineConfig({
   testDir: "./e2e",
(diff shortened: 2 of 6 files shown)


eshaanag:
Rebased this branch onto current main and pushed a follow-up after checking the failed Playwright report. The local-coding fix is still the core change, and the extra CI fixes cover the standalone E2E server missing static assets plus dashboard/public-profile auth edge cases that were breaking the smoke suite. Local validation is green: lint, type-check, the local-coding sync regression test, Node 20 production build, and the full Playwright suite.

Priyanshu-byte-coder (commented):
This PR now has merge conflicts with main due to recent merges. Please rebase onto main and resolve conflicts.

eshaanag:
Rebased onto the latest main and resolved the conflict from recent merges. I kept the local-coding session-cap fix intact and retained the middleware/E2E follow-ups needed for the smoke suite.

Latest validation is green locally: lint, type-check, focused local-coding sync test, Node 20 production build, full Playwright, plus the affected dashboard-widgets Playwright spec after the recent dashboard changes. GitHub CI/build/Playwright are passing now too.

The only remaining red check is the PR Labeler workflow. After the repo added `.github/labeler.yml`, it now fails with `Resource not accessible by integration`, which is the usual fork-PR token permission issue for a label-writing workflow running on `pull_request`. That looks unrelated to this PR code.

PR #1613, opened 2026-05-30
fix(leaderboard): evict deleted users from cached leaderboard results

### LINKED ISSUE #934: [BUG] Leaderboard cache never evicts entries for users who delete their account
## Describe the Bug

The leaderboard in-memory cache stores user entries indefinitely. When a user deletes their account, their entry remains in the leaderboard until the server restarts. This violates user data deletion expectations.

## Steps to Reproduce

1. Add a user to the leaderboard
2. Delete that user's account
3. View the leaderboard — deleted user still appears

## Expected Behavior

Cache entry for a deleted user should be invalidated immediately on account deletion, or the leaderboard fetch should verify user existence before serving cached entries.

## Screenshots / Logs

N/A

## Environment

- OS: Any
- Node version: 18+
- Browser: Any

## Additional Context

The leaderboard route uses TTL-based eviction added in issue #806. The user deletion route should also explicitly purge that user's cache key.

mallya-m:
## What does this PR do?
Fixes deleted users appearing in the leaderboard indefinitely. Cached results are now filtered against current public users in Supabase before serving.

## Related issue
Closes #934

## Changes made
- Added `filterDeletedUsers()` that cross-references cached usernames against current public users in DB
- Applied to both memory cache and distributed cache hits
- Re-ranks remaining entries after filtering
- Also fixes pre-existing build error: missing `bio` field in `public-profile-data.ts`

## How to test
1. Have a user on the leaderboard
2. Set their `is_public = false` in Supabase
3. Leaderboard no longer shows them without server restart

### DIFF
--- a/src/app/api/leaderboard/route.ts
+++ b/src/app/api/leaderboard/route.ts
@@ -19,6 +19,50 @@ import {
 } from "@/lib/upstash-rest";
 
 export const dynamic = "force-dynamic";
+/**
+ * Removes leaderboard entries for users who have deleted their account.
+ * Called before serving any cached payload to ensure data deletion compliance.
+ */
+async function filterDeletedUsers(
+  payload: LeaderboardPayload
+): Promise<LeaderboardPayload> {
+  // Collect all unique usernames across all metric categories
+  const allUsernames = new Set<string>();
+  for (const entries of Object.values(payload.leaders)) {
+    for (const entry of entries) {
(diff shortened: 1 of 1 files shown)


Priyanshu-byte-coder:
Closing — superseded by PR #1616 which includes the same leaderboard changes plus additional improvements to GoalTracker and other components. Please follow the discussion on #1616.

PR #1673, opened 2026-05-30
Fix calculateWeeklyTrend bug by averaging sums (#1215)

### LINKED ISSUE #1215: [Bug]: Math/Logic error in `calculateWeeklyTrend` results in incorrect upward trend for stable activity
Hi team,

I found a calculation bug in `src/lib/coding-activity-insights.ts` inside the `calculateWeeklyTrend` function. 

Right now, `firstHalf` sums up 3 days and `secondHalf` sums up 4 days. If a user has exactly 1 commit every day (perfectly stable activity), `secondHalf` (4) is compared to `firstHalf` (3). This incorrectly shows a 33% "up" trend instead of "stable".

**Fix:**
We just need to divide the sums by the number of days to get the daily average before comparing them:
```typescript
const firstHalfAvg = firstHalfSum / firstHalfDays.length;
const secondHalfAvg = secondHalfSum / secondHalfDays.length;

bhavyapandiya29:
### Description
This PR resolves the math/logic error in the `calculateWeeklyTrend` function within `src/lib/coding-activity-insights.ts`. Previously, the function compared the raw sums of a 4-day period against a 3-day period, incorrectly resulting in an "up" trend for perfectly stable activity. The sums are now correctly divided by the number of days to yield a daily average before performing the trend comparison. 

Fixes #1215.

### DIFF
--- a/src/lib/coding-activity-insights.ts
+++ b/src/lib/coding-activity-insights.ts
@@ -228,13 +228,13 @@ function calculateWeeklyTrend(
   direction: "up" | "down" | "stable";
   percentage: number;
 } {
-  const firstHalf = dayCounts
-    .slice(0, 3)
-    .reduce((sum, day) => sum + day.count, 0);
+  const firstHalfDays = dayCounts.slice(0, 3);
+  const firstHalfSum = firstHalfDays.reduce((sum, day) => sum + day.count, 0);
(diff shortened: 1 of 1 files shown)


bhavyapandiya29:
Hi @Priyanshu-byte-coder I have made a PR for this issue. Kindly review and merge it.

PR #1719, opened 2026-05-31
"feat: validate package subpath exports in check-deps"

### LINKED ISSUE #1712: check-deps.js does not validate package subpath exports
Problem

The current dependency checker verifies that imported packages exist in package.json, but it does not verify whether imported package subpaths are actually exported by the dependency.

Example:

import helper from "some-package/internal/helper";

If "some-package" exists in package.json, the check passes even if "internal/helper" is not exposed through the package's exports field.

Expected Behavior

The checker should validate package subpath imports against the dependency's exports configuration and report invalid imports.

Benefit

This would help catch invalid imports earlier and prevent build-time or runtime errors.

namrarafique93-del:
Summary

Adds validation for package subpath imports in "check-deps.js".

Previously, the dependency checker only verified that the base package existed in "package.json". This change adds an additional check for package subpath imports and reports imports that are not exposed through the package's "exports" field.

Example:

import helper from "some-package/internal/helper";

If "some-package/internal/helper" is not exported by the package, the checker will now report it.

Changes Made

- Added "isValidPackageSubpath()" helper function
- Added validation for package subpath imports
- Updated dependency checking logic in "collectMissingDeps()"
- Reports invalid package subpath imports as missing dependencies

Why

This helps catch invalid imports earlier and prevents build-time or runtime failures caused by importing private package internals.

Testing

- Verified existing dependency checks continue to work
- Verified package subpath imports are validated before being accepted
- Confirmed no changes to relative imports, aliases, framework aliases, or Node built-in modules

Closes #1712

### DIFF
--- a/scripts/check-deps.js
+++ b/scripts/check-deps.js
@@ -94,6 +94,40 @@ function loadInternalAliases(rootDir) {
 
   return aliases;
 }
+function isValidPackageSubpath(pkgName, mod, cwd) {
+  try {
+    const pkgJsonPath = require.resolve(
+      `${pkgName}/package.json`,
+      { paths: [cwd] }
+    );
+
+    const pkgJson = JSON.parse(
+      fs.readFileSync(pkgJsonPath, "utf8")
+    );
+
+    const exportsField = pkgJson.exports;
+
+    if (!exportsField) return true;
+
+    const subpath = mod.slice(pkgName.length);
+
+    if (!subpath) return true;
+
+    const exportKey =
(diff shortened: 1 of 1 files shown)


namrarafique93-del:
Hii @Priyanshu-byte-coder 
The failing GitHub Action appears unrelated to this change.

The failures are coming from Playwright smoke tests and application environment setup ("sharp" and Supabase configuration), not from "scripts/check-deps.js".

Please let me know if there is a specific test suite I should run or if additional changes are expected for this issue 
Thank you !!

PR #1730, opened 2026-05-31
fix: show native share button on desktop browsers supporting Web Share API

### LINKED ISSUE #1717: [BUG] Native Share Button Not Available on Desktop Browsers Supporting Web Share API
## Description

The native Share button in `ShareProfileSection.tsx` is currently displayed only when the device uses a coarse pointer (typically mobile devices). However, several modern desktop browsers also support the Web Share API (`navigator.share()`), causing the Share button to be unnecessarily hidden on those platforms.

### File

`src/components/ShareProfileSection.tsx`

### Current Code

```tsx
setCanUseNativeShare(
  typeof navigator !== "undefined" &&
    "share" in navigator &&
    typeof window !== "undefined" &&
    window.matchMedia("(pointer: coarse)").matches
);
```

### Expected Behavior

The Share button should be visible whenever the browser supports the Web Share API, regardless of whether the user is on a mobile or desktop device.

### Actual Behavior

On desktop browsers that support `navigator.share()`, the Share button remains hidden because of the additional coarse pointer check.

### Steps to Reproduce

1. Open DevTrack in a desktop browser that supports the Web Share API.
2. Navigate to the profile sharing section.
3. Observe that the Share button is not displayed.
4. Only platform-specific sharing options such as X and LinkedIn are available.

Kokila-chandrakar:
## Description

This PR fixes an issue where the native Share button was only displayed on devices with a coarse pointer (typically mobile devices).

The previous implementation required both Web Share API support and a coarse pointer device, which prevented the Share button from appearing on desktop browsers that already support `navigator.share()`.

### Changes Made

* Removed the `pointer: coarse` device check.
* Enabled the Share button whenever the browser supports the Web Share API.
* Ensured a consistent sharing experience across supported mobile and desktop browsers.

### Before

The Share button was hidden on desktop browsers even when `navigator.share()` was available.

### After

The Share button is displayed on any browser that supports the Web Share API, regardless of device type.

### Testing

* Verified that the Share button appears when `navigator.share()` is supported.
* Confirmed that existing sharing functionality continues to work as expected.

Fixes #1717

### DIFF
--- a/src/components/ShareProfileSection.tsx
+++ b/src/components/ShareProfileSection.tsx
@@ -20,9 +20,7 @@ export default function ShareProfileSection({
   useEffect(() => {
     setCanUseNativeShare(
       typeof navigator !== "undefined" &&
-        "share" in navigator &&
-        typeof window !== "undefined" &&
-        window.matchMedia("(pointer: coarse)").matches
+        "share" in navigator
     );
   }, []);
 


Kokila-chandrakar:
Hi @Priyanshu-byte-coder,

The implementation is complete and the PR is ready for review. The failing check appears to be related to the repository labeler workflow permissions (`Resource not accessible by integration`) rather than the code changes.

Thank you!

PR #1765, opened 2026-05-31
fix: store ai_insights with users.id and add FK + cascade delete

### LINKED ISSUE #1750: [BUG] AI insights records become orphaned because user references are stored using GitHub IDs instead of application user IDs
## Bug Summary

The AI insights system stores GitHub numeric identifiers instead of application user identifiers.

Because no foreign-key relationship exists, AI insight records survive account deletion and become orphaned.

## Describe the bug

The AI insights table stores:

session.githubId

instead of:

users.id

As a result:

- foreign-key constraints cannot be enforced
- cascade deletes cannot operate
- records become disconnected from user accounts

Deleting a user account leaves associated AI insights behind.

## To Reproduce

1. Create AI insight records.
2. Delete the associated user account.
3. Query ai_insights.
4. Observe that the records remain.
5. Observe that they no longer correspond to a valid user row.

## Expected behavior

User-owned AI insights should be deleted when the user account is removed.

## Actual behavior

AI insights become orphaned.

## Additional context

Affected files:

- src/app/api/ai-insights/route.ts
- supabase/schema.sql

Potential impact:

- data retention issues
- privacy concerns
- orphaned records
- broken referential integrity

Ridanshi:
Closes #1750

## Problem

`GET /api/ai-insights` stored insight rows using `session.githubId` as `user_id`:

```ts
const userId = session.githubId;  // e.g. "12345678" — GitHub numeric ID
// ...
await supabaseAdmin.from("ai_insights").upsert({ user_id: userId, ... });
```

`session.githubId` is the **GitHub numeric account ID**, not the application UUID (`users.id`). The `ai_insights` table had no foreign-key constraint pointing at `users`, so:

1. **Orphaned records** — when a user deleted their account, all their `ai_insights` rows remained forever. The account-deletion handler (`data-export/route.ts`) did not include `ai_insights` in its explicit deletion list, and there was no `ON DELETE CASCADE` to fill in.

2. **No referential integrity** — inserts could reference any arbitrary string as `user_id` with no guarantee that a matching user existed.

3. **Broken cascade** — even if the cascade were to fire in the future, it would compare against `users.id` (UUID), not `users.github_id` (numeric string), so it would never match existing rows.

## Root cause (three locations)

| Location | Problem |
|---|---|
| `src/app/api/ai-insights/route.ts` line 51 | `const userId = session.githubId` — uses wrong identity |
| `src/app/api/user/data-export/route.ts` — `tablesToDelete` array | `ai_insights` not listed; rows survive account deletion |
| `supabase/schema.sql` — `ai_insights` table | No FK, no cascade |

### `src/app/api/ai-insights/route.ts`

- Call `resolveAppUser(session.githubId, session.githubLogin)` to obtain `user.id` (the stable UUID) and use it as `user_id` throughout.
- Return 404 if the user cannot be resolved.
- Add `insight_type` validation against the allowlist `('weekly_summary' | 'pattern' | 'recommendation')` that mirrors the DB `CHECK` constraint. Unrecognised types now return 400 instead of a Supabase constraint-violation 500.

### `src/app/api/user/data-export/route.ts`

- Prepend `"ai_insights"` to `tablesToDelete` so explicit account deletion always removes insight rows — even before the FK migration is applied.

### `supabase/migrations/20260531000000_fix_ai_insights_user_id.sql`

Three steps inside a transaction:

1. **Remap** existing `ai_insights.user_id` values from `github_id` strings to the corresponding `users.id` UUIDs (via `JOIN users ON users.github_id = ai_insights.user_id`).
2. **Delete** any rows that cannot be matched — these are already orphaned records belonging to deleted accounts.
3. **Add FK** `FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE`, enforcing integrity going forward and enabling automatic cleanup on account deletion.

## Identity chain (after fix)

```
GitHub OAuth sign-in  →  profile.id  →  users.github_id  (immutable)
                                     →  users.id (UUID)  ← ai_insights.user_id now points here
```

## Tests

**`test/ai-insights-ownership.test.ts`** — 13 new tests:

| Category | Cases |
|---|---|
| Auth guards | no session → 401, unresolved user → 404 |
| Ownership | `resolveAppUser` called with session credentials; upsert uses UUID; cache query uses UUID |
| Regression #1750 | no DB operation receives the raw `githubId` as `user_id` |
| Cache hit | returns `{ cached: true, data }` |
| `insight_type` validation | invalid → 400; `weekly_summary`, `pattern`, `recommendation` → 200 |
| Conflict target | `onConflict: "user_id,insight_type"` present |
| Cache miss shape | `{ data, cached: false }` |

All 13 pass. The single pre-existing failure in `test/dateUtils.test.ts` (timezone boundary) is unrelated to this change.

### DIFF
--- a/src/app/api/ai-insights/route.ts
+++ b/src/app/api/ai-insights/route.ts
@@ -2,17 +2,28 @@ import { NextResponse } from "next/server";
 import { getServerSession } from "next-auth";
 import { authOptions } from "@/lib/auth";
 import { supabaseAdmin } from "@/lib/supabase";
+import { resolveAppUser } from "@/lib/resolve-user";
 import { weeklyProductivityPrompt } from "@/lib/ai-prompts";
 import {
   analyzePatterns,
   computeTrends,
   DeveloperMetrics,
 } from "@/lib/ai-mentor";
+
 const aiInsightsRateLimit = new Map<
   string,
   { count: number; resetTime: number }
 >();
+
 export const dynamic = "force-dynamic";
+
+const VALID_INSIGHT_TYPES = new Set([
+  "weekly_summary",
+  "pattern",
+  "recommendation",
+] as const);
+
+type InsightType = "weekly_summary" | "pattern" | "recommendation";
 
 interface ContributionsApiResponse {
   data?: Record<string, number>;
@@ -48,7 +59,18 @@ export async function GET(request: Request) {
     return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
   }
 
-  const userId = session.githubId;
+  // Resolve the application user so that ai_insights rows are keyed on
+  // users.id (a stable UUID with a foreign-key constraint) rather than on
+  // session.githubId (a mutable external identifier with no FK). Using
+  // users.id ensures rows are removed via ON DELETE CASCADE when the account
+  // is deleted and prevents the orphaned-record scenario described in #1750.
+  const user = await resolveAppUser(session.githubId, session.githubLogin);
+  if (!user) {
+    return NextResponse.json({ error: "User not found" }, { status: 404 });
+  }
+
+  const userId = user.id;
+
   const currentTime = Date.now();
   const WINDOW_MS = 60 * 60 * 1000;
   const MAX_REQUESTS = 5;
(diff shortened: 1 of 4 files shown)


Priyanshu-byte-coder:
This PR has developed merge conflicts after recent merges to `main`. Please rebase onto the latest `main` branch, resolve all conflicts, and force-push. Once the PR is conflict-free it will be reviewed for merging.

PR #1841, opened 2026-06-01
chore: add prominent security warning to .env.example

### LINKED ISSUE #1828: [SECURITY/CHORE] Add strict warning to .env.example and rotate previously leaked keys
"To prevent this from happening again, I would like to be assigned to this issue so I can add a prominent security warning to the top of .env.example telling contributors not to place real keys in that tracked file."

Akshita-2307:
Adds a prominent security banner at the top of .env.example warning contributors not to place real secrets in the tracked template file. Instructs to use .env.local instead and to rotate any leaked keys.

Closes #1828


Priyanshu-byte-coder:
Thanks! This was already addressed in commit 73682c3. Closing.

PR #1869, opened 2026-06-02
fix: No rate limiting on the API-key-authenticated local coding sync 

### LINKED ISSUE #1690: [Bug] No rate limiting on the API-key-authenticated local coding sync endpoint
## Summary

The `POST /api/local-coding/sync` endpoint authenticates via API key but applies no rate limiting. Every leaderboard, badge, and public-profile route in the codebase has an explicit rate limiter, but this write endpoint is unprotected.

## Affected File

`src/app/api/local-coding/sync/route.ts`

```typescript
export async function POST(request: Request) {
  // No rateLimit(...) call here - compare with /api/leaderboard/route.ts
  // which correctly applies rateLimitMiddleware
  const apiKey = request.headers.get('x-api-key');
  // ... authenticate key, then bulk upsert sessions
}
```

## Impact

Any holder of a valid API key (obtained legitimately or through the hash bypass described in the companion issue) can continuously hammer this endpoint:
- Each request triggers a Supabase count query and a batched upsert RPC.
- At high frequency this exhausts database connections, Supabase row limits, and serverless function invocations.
- The VSCode extension that calls this endpoint runs automatically in the background; a bug in the extension could trigger an unintended burst.

## Suggested Fix

Apply the same `rateLimitMiddleware` pattern used by `/api/leaderboard/route.ts`, keyed on the hashed API key rather than the user's IP, to allow legitimate high-frequency syncs while preventing abuse.

## Severity

Medium

Aditya8369:
Rate limiting added to POST /api/local-coding/sync keyed by the sha256 hash of the bearer API key (fixed window: 20 requests per 60s). When exceeded, the endpoint now returns 429 with { error: "Rate limit exceeded" } and a Retry-After header.

Updated files:

src/app/api/local-coding/sync/route.ts (added in-memory + Upstash-backed limiter; applied only to POST)
test/local-coding-sync.test.ts (added test asserting 21st POST returns 429; mocked Upstash config to force in-memory)

closes #1690

### DIFF
--- a/src/app/api/local-coding/sync/route.ts
+++ b/src/app/api/local-coding/sync/route.ts
@@ -1,6 +1,8 @@
 import { NextRequest } from "next/server";
 import { supabaseAdmin } from "@/lib/supabase";
 import { createHash } from "crypto";
+import { getUpstashConfig, upstashRateLimitFixedWindow } from "@/lib/upstash-rest";
+
 
 export const dynamic = "force-dynamic";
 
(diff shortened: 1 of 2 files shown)


Aditya8369:
@Priyanshu-byte-coder please check this one

Aditya8369:
@Priyanshu-byte-coder please check

PR #1936, opened 2026-06-03
fix: guard contribution graph zero commits

### LINKED ISSUE #1889: [BUG] ContributionGraph shows NaN when user has zero commits in selected range
## Describe the bug
When a user has zero commits in the currently selected time range (7d / 14d / 30d / 90d), the ContributionGraph widget renders `NaN` in the total count label instead of `0`.

## Steps to reproduce
1. Sign in to DevTrack
2. Go to the dashboard
3. In the Contribution Graph, select the **7d** range immediately after creating a new account (no commits yet)
4. Observe the total count label showing `NaN commits`

## Expected behaviour
Total count should show **0 commits** when there are no commits in the selected range.

## Suggested fix
Add a null/zero guard in the count calculation:
```ts
const total = commits.reduce((sum, c) => sum + (c.count ?? 0), 0) || 0;
```

## Environment
- Browser: Any
- Auth: GitHub OAuth

saurabhhhcodes:
Closes #1889

## Summary
- Normalize contribution counts before chart/compare rendering so undefined, non-finite, or negative values become 0.
- Show a finite total count in the ContributionGraph header for zero-commit ranges.
- Add a regression test that verifies empty/undefined contribution ranges render `0 commits` instead of `NaN commits`.

## Validation
- `npx vitest run test/ContributionGraph.test.ts`
- `npm run type-check`
- `npm run lint`
- `git diff --check`

Note: `npm run lint` passes with existing warnings in unrelated files:
- `src/app/api/og/user/route.tsx`
- `src/components/DailyNoteWidget.tsx`

### DIFF
--- a/src/components/ContributionGraph.tsx
+++ b/src/components/ContributionGraph.tsx
@@ -58,6 +58,29 @@ const charts: { key: ViewMode; label: string }[] = [
   { key: "area", label: "Area" },
 ];
 
+function normalizeCommitCount(count: unknown): number {
+  const numericCount = Number(count ?? 0);
+
+  if (!Number.isFinite(numericCount) || numericCount <= 0) {
+    return 0;
+  }
+
+  return numericCount;
+}
+
+function normalizeContributionData(data: Record<string, number>): DayData[] {
+  return Object.entries(data)
+    .sort(([a], [b]) => a.localeCompare(b))
+    .map(([day, commits]) => ({
+      day,
+      commits: normalizeCommitCount(commits),
+    }));
+}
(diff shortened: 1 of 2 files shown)


saurabhhhcodes:
Thanks for merging this so quickly. One small scoring follow-up: I can see the PR has `level:intermediate`, `quality:clean`, `type:bug`, and `type:testing`, but it does not currently show `gssoc:approved` after merge. Could you please add `gssoc:approved` as well if this is accepted for GSSoC scoring?

PR #1969, opened 2026-06-03
fix: persist recurring goal history on reset

### LINKED ISSUE #1847: [BUG] Recurring goals silently erase completion history on period reset

nishupr:
## Summary
Fixes recurring goal resets so previous weekly/monthly period results are persisted before progress is reset.

## Changes
- Added a `goal_history` table with period result fields.
- Records the previous period result before resetting recurring goals in `GET /api/goals`.
- Returns the latest previous period as `last_period` with each goal.
- Displays a small “Last period” result on recurring goal cards.
- Includes goal history in JSON data exports.
- Cleans up `goal_history` during account deletion.
- Updated GoalTracker tests for the new response shape.

## Verification
- `npm run type-check`
- `npm run build`
- `npx vitest run --config ./vitest.config.ts test/GoalTracker.test.ts`

Closes #1847

### DIFF
--- a/src/app/api/goals/route.ts
+++ b/src/app/api/goals/route.ts
@@ -20,6 +20,15 @@ interface Goal {
   created_at: string;
 }
 
+interface GoalHistory {
+  goal_id: string;
+  period_start: string;
+  period_end: string;
+  target: number;
+  achieved: number;
+  completed: boolean;
+}
+
 type Recurrence = "none" | "weekly" | "monthly";
 
 const VALID_RECURRENCES = ["none", "weekly", "monthly"] as const;
@@ -45,6 +54,10 @@ function getPeriodStart(recurrence: Recurrence): string {
     return new Date(Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), 1)).toISOString();
   }
   return new Date(0).toISOString(); // 'none' never resets
+}
+
+function getPreviousPeriodEnd(periodStart: Date): string {
+  return new Date(periodStart.getTime() - 1).toISOString();
 }
 
 export async function GET() {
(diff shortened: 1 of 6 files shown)


nishupr:
hey @Priyanshu-byte-coder could you please add level label to this PR too?

PR #1975, opened 2026-06-03
fix: persist daily focus goals in database

### LINKED ISSUE #1848: [BUG] "Today's Focus" goal is stored only in localStorage — lost on new device, incognito, or browser clear
## Description

The `TodayFocusHero` widget on the dashboard lets users set a daily focus goal. 
However, it stores that goal exclusively in `localStorage` using a date-keyed entry 
(`devtrack_today_goal_YYYY-MM-DD`). It is never saved to the database or associated 
with the user's account in any way.

## Steps to Reproduce

1. Sign in and set a "Today's Focus" goal on the dashboard
2. Open the dashboard in a different browser, incognito window, or on your phone
3. The goal is completely absent — the widget shows the empty input prompt again
4. Clearing browser storage or switching devices has the same effect

## Expected Behaviour

Since DevTrack already has a Supabase backend and a goals system, the daily focus 
should be persisted server-side per user, so it survives across devices and sessions.

## Actual Behaviour

In `TodayFocusHero.tsx`, all reads and writes go directly to `window.localStorage`:

```ts
// Read
const storedGoal = window.localStorage.getItem(nextKey)?.trim() ?? "";

// Write
window.localStorage.setItem(todayKey, trimmedGoal);

// Delete
window.localStorage.removeItem(todayKey);
```

There is no API call, no Supabase write, and no association with the authenticated user.

## Additional Impact

- The daily focus is also excluded from the **data export** (JSON dump), since it never 
  touches the database
- Users who clear browser storage as part of routine maintenance silently lose their focus history
- The "Today's Focus" feature is prominently placed at the top of the dashboard, 
  making it one of the most visible features — yet it behaves like a sticky note 
  tied to a single browser tab

## Suggested Fix

1. Add a `daily_focus` table (or a `daily_notes` column on an existing user table) 
   with columns: `user_id`, `date` (YYYY-MM-DD), `goal_text`
2. Add `GET /api/daily-focus?date=` and `POST /api/daily-focus` route handlers
3. In `TodayFocusHero.tsx`, replace `localStorage` reads/writes with `fetch()` calls 
   to those endpoints
4. Keep a local optimistic update for instant UI feedback while the request is in-flight

## Files Involved

- `src/components/TodayFocusHero.tsx` — entire storage layer needs replacing
- `src/app/api/` — new `daily-focus/route.ts` needed
- `supabase/migrations/` — new migration for the `daily_focus` table

nishupr:
Fixes #1848 

## Description

This PR resolves the issue where the "Today's Focus" goal on the dashboard was only stored in localStorage, causing users to lose their goals when switching devices, using incognito mode, or clearing browser data.

To fix this, the daily focus goals are now persisted in the Supabase database.

## Changes Made

- Database Migration: Added a new daily_focus table (with user_id, date, and goal_text columns) and Row Level Security (RLS) policies to ensure users can only access their own goals.
- Backend Sync API: Created a new set of API endpoints (/api/daily-focus) to handle GET, POST (upsert), and DELETE requests for the daily goals.
- Frontend Refactoring: Refactored TodayFocusHero.tsx to communicate with the new backend API. It still utilizes localStorage as a fallback to maintain the instant, optimistic UI feedback, but now safely syncs state with the server in the background.

### DIFF
--- a/src/app/api/daily-focus/route.ts
+++ b/src/app/api/daily-focus/route.ts
@@ -1,0 +1,115 @@
+import { NextResponse, NextRequest } from "next/server";
+import { supabaseAdmin } from "@/lib/supabase";
+import { getToken } from "next-auth/jwt";
+
+export async function GET(req: NextRequest) {
+  try {
+    const token = await getToken({
+      req,
+      secret: process.env.NEXTAUTH_SECRET,
+    });
+
+    const userId = token?.githubId;
+
+    if (!userId) {
+      return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
+    }
+
+    const { searchParams } = new URL(req.url);
+    let date = searchParams.get("date");
+
+    if (!date) {
+      date = new Date().toISOString().split("T")[0];
+    }
+
+    const { data } = await supabaseAdmin
+      .from("daily_focus")
+      .select("*")
+      .eq("user_id", userId)
+      .eq("date", date)
+      .single();
+
+    return NextResponse.json({
+      goal: data?.goal_text || "",
+    });
+  } catch (error) {
+    return NextResponse.json({ error: "Something went wrong" }, { status: 500 });
+  }
+}
+
+export async function POST(req: NextRequest) {
+  try {
+    const token = await getToken({
+      req,
+      secret: process.env.NEXTAUTH_SECRET,
+    });
+
+    const userId = token?.githubId;
+
+    if (!userId) {
+      return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
+    }
+
+    const body = await req.json();
+    const { goal_text, date } = body;
+
+    if (!goal_text || !goal_text.trim()) {
+      return NextResponse.json({ error: "Goal cannot be empty" }, { status: 400 });
+    }
+
+    const targetDate = date || new Date().toISOString().split("T")[0];
+
+    const { data, error } = await supabaseAdmin
+      .from("daily_focus")
+      .upsert(
+        { user_id: userId, date: targetDate, goal_text: goal_text.trim() },
+        { onConflict: "user_id,date" }
+      )
+      .select()
+      .single();
+
+    if (error) {
+      return NextResponse.json({ error: "Failed to save goal" }, { status: 500 });
+    }
+
+    return NextResponse.json(data);
+  } catch (error) {
+    return NextResponse.json({ error: "Something went wrong" }, { status: 500 });
+  }
+}
+
+export async function DELETE(req: NextRequest) {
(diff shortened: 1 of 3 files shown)


Priyanshu-byte-coder:
The implementation has a **user ID mismatch bug** that would break in production.

`token?.githubId` is the raw GitHub integer ID (e.g. `"123456"`) but `daily_focus.user_id` references `users(id)` which is a UUID. The FK insert would fail.

Every other API in the project calls `resolveAppUser` to bridge that gap:

```ts
// Replace this:
const userId = token?.githubId;

// With this pattern (used in /api/goals, /api/user, etc.):
import { getServerSession } from "next-auth";
import { authOptions } from "@/lib/auth";
import { resolveAppUser } from "@/lib/resolve-user";

const session = await getServerSession(authOptions);
if (!session?.githubId) return NextResponse.json({ error: "Unauthorized" }, { status: 401 });

const user = await resolveAppUser(session.githubId, session.githubLogin);
if (!user) return NextResponse.json({ error: "User not found" }, { status: 404 });

const userId = user.id;  // ← UUID matching users(id)
```

Also please rebase on `main` to pick up recent changes. Once those are fixed this is ready to merge.

nishupr:
> The implementation has a **user ID mismatch bug** that would break in production.
> 
> `token?.githubId` is the raw GitHub integer ID (e.g. `"123456"`) but `daily_focus.user_id` references `users(id)` which is a UUID. The FK insert would fail.
> 
> Every other API in the project calls `resolveAppUser` to bridge that gap:
> 
> ```ts
> // Replace this:
> const userId = token?.githubId;
> 
> // With this pattern (used in /api/goals, /api/user, etc.):
> import { getServerSession } from "next-auth";
> import { authOptions } from "@/lib/auth";
> import { resolveAppUser } from "@/lib/resolve-user";
> 
> const session = await getServerSession(authOptions);
> if (!session?.githubId) return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
> 
> const user = await resolveAppUser(session.githubId, session.githubLogin);
> if (!user) return NextResponse.json({ error: "User not found" }, { status: 404 });
> 
> const userId = user.id;  // ← UUID matching users(id)
> ```
> 
> Also please rebase on `main` to pick up recent changes. Once those are fixed this is ready to merge.

@Priyanshu-byte-coder 
Thanks for catching that. Fixed the `daily_focus.user_id` mismatch by resolving the session GitHub ID through `resolveAppUser` and using the app user UUID for GET/POST/DELETE. Also rebased on latest `main`.

Validation:
- `npm.cmd run type-check`
- `npm.cmd run build`

PR #2004, opened 2026-06-04
fix: paginate weekly-digest cron to prevent serverless timeout

### LINKED ISSUE #1852: Bug: weekly-digest cron fetches all opted-in users with no limit, causing serverless timeout when user count is large
## What happened?

`src/app/api/cron/weekly-digest/route.ts` fetches every opted-in user in a single unbounded Supabase query and then sends emails sequentially in a `for` loop within one serverless function invocation:

```typescript
// Unbounded query - loads ALL opted-in users at once
    .from("users")
    .select("github_login, email")
    .eq("weekly_digest_opt_in", true)
    .not("email", "is", null);

// Sequential loop with no batching or timeout guard
for (const user of users) {
    if (!user.email) continue;
    // ... await email send per user ...
}
```

Vercel's default serverless function timeout is 10 seconds (60 seconds on Pro). Each email send is an awaited network request. With 100 opted-in users and an average email API latency of 300ms, the function runs for approximately 30 seconds. At 200+ users it reliably times out.

When the function times out, partially-sent digests produce inconsistent delivery: some users receive the email, others do not, with no way to determine who was reached and no retry mechanism.

## Steps to Reproduce

1. Opt in 100 or more users via `weekly_digest_opt_in = true` in the database.
2. Trigger `GET /api/cron/weekly-digest` with the correct `Authorization: Bearer <CRON_SECRET>` header.
3. Observe the function timeout before completing the full user list.
4. Check delivery logs - only a subset of opted-in users received the email.

## Expected Behavior

The cron handler should process users in paginated batches (e.g. 50 at a time) so that each invocation completes within the serverless timeout window. Or it should enqueue jobs to a queue/worker system for background processing.

## Actual Behavior

All opted-in users are loaded into memory in one query and emailed sequentially inside a single function invocation. With a growing user base, this will timeout and silently skip users.

## Root Cause Analysis

There is no `.limit()` or `.range()` clause on the Supabase query, and no batching around the email send loop. The implementation works correctly at small scale (10-20 users) but degrades to a partial-delivery failure at production scale.

## Suggested Fix

Add a `.limit()` clause and process in chunks across multiple invocations, or use the existing queue infrastructure if one is available:

```typescript
const PAGE_SIZE = 50;
let page = 0;
let hasMore = true;

while (hasMore) {
    const { data: users } = await supabaseAdmin
        .from("users")
        .select("github_login, email")
        .eq("weekly_digest_opt_in", true)
        .not("email", "is", null)
        .range(page * PAGE_SIZE, (page + 1) * PAGE_SIZE - 1);

    if (!users || users.length < PAGE_SIZE) hasMore = false;
    // ... send emails for this page ...
    page++;
}
```

## Affected File

`src/app/api/cron/weekly-digest/route.ts`

## Environment

- Runtime: Next.js App Router (serverless)
- Deployment target: Vercel (10s/60s timeout)

## Checklist

- [x] Searched existing issues, not a duplicate
- [x] Read CONTRIBUTING.md guidelines
- [x] Provided clear reproduction steps
- [x] Described expected vs. actual clearly
- [x] No em dashes or double hyphens
- [x] Repository verified as GSSoC

Kokila-chandrakar:
## Problem
Unbounded Supabase query loads all opted-in users at once and emails 
them sequentially in a single serverless invocation. At 100+ users 
this exceeds Vercel's 10s timeout causing partial delivery with no 
retry mechanism.

## Changes
- Replace unbounded query with paginated `.range()` loop (PAGE_SIZE=50)
- Send emails in parallel per page using `Promise.allSettled()`
- Extract `sendDigestEmail()` helper with Resend API error handling
- Return `sentCount` and `errorCount` in response for observability

## Affected File
`src/app/api/cron/weekly-digest/route.ts`

## Testing
1. Opt in 100+ users via `weekly_digest_opt_in = true`
2. Trigger `GET /api/cron/weekly-digest` with correct `Authorization` header
3. Verify all users receive email and function completes within timeout

##close 

fix #1852


Priyanshu-byte-coder:
Closing in favor of #2068 which landed first and provides a complete digest implementation (proper email template, unsubscribe link, metrics, pagination with batch-of-5). The pagination fix from this PR is covered there. Thanks for the contribution!

PR #2080, opened 2026-06-06
fix(auth): prevent API key hash replay through dual-column lookup

### LINKED ISSUE #1689: [Bug][Security] API key hash written to both api_key and api_key_hash columns, bypassing cryptographic protection
## Summary

When a new API key is created in `src/app/api/local-coding/keys/route.ts`, the same hash value is written to both `api_key` and `api_key_hash` columns. The authentication query in `sync/route.ts` then searches both columns, meaning the stored hash itself can be used as a valid credential, defeating any security benefit of hashing.

## Affected Files

**Creation** (`src/app/api/local-coding/keys/route.ts`):
```typescript
await supabase.from('api_keys').insert({
  api_key: apiKeyHash,      // hash stored here
  api_key_hash: apiKeyHash, // and again here
  user_id: userId,
});
```

**Authentication** (`src/app/api/local-coding/sync/route.ts`):
```typescript
const keyHashFilter = `api_key_hash.eq.${keyHash},api_key.eq.${keyHash}`;
const { data } = await supabase.from('api_keys')
  .select('*')
  .or(keyHashFilter);
```

## Impact

- An attacker who gains read access to the `api_keys` table (e.g., through a SQL injection elsewhere, a misconfigured RLS policy, or a database backup leak) can directly use the stored `api_key` column value as the lookup token, without needing to reverse the hash.
- The dual-write also suggests the `api_key` column may still contain raw keys for rows created before any hash migration, making the fallback actively dangerous.

## Suggested Fix

1. Only write to `api_key_hash`. Remove the `api_key` column or migrate it to store only a display prefix (e.g., first 8 characters).
2. Remove `api_key.eq.${keyHash}` from the OR filter so authentication only queries the hash column.

## Severity

High

Ridanshi:
Closes #1689

## Root cause

Key creation wrote the SHA-256 hash to **both** `api_key` and `api_key_hash`. Authentication accepted a match on either column via an OR filter (`api_key_hash.eq.${hash},api_key.eq.${hash}`). This made `api_key` — a column treated as a display field — carry real authentication weight. Any value readable from the database could not be trivially replayed due to the double-hash property, but the dual-column design violated the principle that credential-derived values must be confined to explicitly secured columns and created an unnecessary attack surface.

## Security impact

- Removed the `api_key` fallback from authentication entirely. Only `api_key_hash` is queried.
- The `api_key` column no longer holds a hash — it now holds only an 8-character display prefix, which has no authentication value.
- A stolen `api_key_hash` value cannot authenticate: the route hashes the incoming bearer token before lookup, so `sha256(stolenHash) ≠ stolenHash`.

## Storage changes

**Before:** `api_key = sha256(rawKey)`, `api_key_hash = sha256(rawKey)`

**After:** `api_key = rawKey.slice(0, 8)` (display prefix), `api_key_hash = sha256(rawKey)`

## Authentication changes

**Before:**
```typescript
.or(`api_key_hash.eq.${keyHash},api_key.eq.${keyHash}`)
```

**After:**
```typescript
.eq("api_key_hash", keyHash)
```

The `last_used_at` update uses the same single-column filter.

## Migration

`supabase/migrations/20260603000000_backfill_api_key_hash.sql` handles three legacy row variants:

| Variant | api_key | api_key_hash | Action |
|---------|---------|--------------|--------|
| A (pre-hash-column) | `sha256(key)` | NULL | Promote hash to `api_key_hash`; replace `api_key` with id prefix |
| B (dual-write era) | `sha256(key)` | `sha256(key)` | Replace `api_key` with id prefix |
| C (current) | 8-char prefix | `sha256(key)` | No action needed |

Existing keys continue to authenticate after migration. No keys are invalidated.

## Tests added

- `test/local-coding-auth-regression.test.ts` — rewritten for #1689: verifies prefix-only storage, hash-only auth, stolen-hash rejection, no `api_key` fallback, and hash-function consistency
- `test/local-coding-keys.test.ts` — updated: verifies `api_key` receives prefix, not hash
- `test/local-coding-sync.test.ts` — updated: mock chains updated from `.or()` to `.eq()`, auth test verifies only `api_key_hash` is consulted

## Verification

```
pnpm lint     → ✓ No ESLint warnings or errors
pnpm typecheck → ✓ No errors in changed files (pre-existing unrelated errors in other feature-branch test files)
pnpm test     → ✓ 25/25 tests pass (local-coding-auth-regression, local-coding-keys, local-coding-sync)
```

### DIFF
--- a/supabase/migrations/20260603000000_backfill_api_key_hash.sql
+++ b/supabase/migrations/20260603000000_backfill_api_key_hash.sql
@@ -1,0 +1,37 @@
+-- Safe migration to enforce canonical hash storage for local_coding_api_keys.
+--
+-- Schema history:
+--   20260521000000_add_local_coding_tables.sql  -- api_key column only (stored hashes)
+--   20260522000000_add_api_key_hash_column.sql  -- api_key_hash column added (nullable)
+--
+-- Three row variants may exist in production:
+--
+--   A. api_key = sha256(key),  api_key_hash = NULL
+--      Created before api_key_hash was introduced. The hash lives only in
+--      api_key.  Promote it to api_key_hash so authentication can use the
+--      canonical column exclusively, then replace api_key with a display
+--      prefix derived from the row id.
+--
+--   B. api_key = sha256(key),  api_key_hash = sha256(key)
+--      Created by the dual-write "fix" for #1748. api_key_hash is already
+--      correct; api_key just needs to be replaced with a non-sensitive prefix.
+--
+--   C. api_key = prefix (8 chars), api_key_hash = sha256(key)
+--      Created by the current code after issue #2038.  Nothing to do.
+--
+-- A SHA-256 hex digest is exactly 64 lowercase hex characters.  That pattern
+-- distinguishes variants A/B from variant C.
+
+-- Step 1: promote hash to api_key_hash for rows that have no hash column yet.
+UPDATE local_coding_api_keys
+SET api_key_hash = api_key
+WHERE api_key_hash IS NULL
+  AND api_key ~ '^[a-f0-9]{64}$';
+
+-- Step 2: replace the hash stored in api_key with a display-safe prefix so
+-- the column no longer holds any credential-derived value.  The prefix is the
+-- first 8 characters of the row's own primary-key UUID, which is unique and
+-- non-sensitive.
+UPDATE local_coding_api_keys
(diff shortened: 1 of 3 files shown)


Priyanshu-byte-coder:
This PR has conflicts with main. Please rebase:
```
git fetch origin && git rebase origin/main
```
Resolve any conflicts, then push. We'll review and merge once CI passes.

PR #2081, opened 2026-06-06
feat(auth): add GitHub App installation token service layer

### LINKED ISSUE #238: [FEAT] Add GitHub App installation mode — no user OAuth token required
## Summary
DevTrack currently requires users to OAuth and uses their personal token for API calls. This hits the user's personal rate limit and requires broad scope. A GitHub App installation would use a bot token with tighter permissions and higher rate limits.

## Expected Behaviour
- Users can install the DevTrack GitHub App on their account/org
- App generates installation access tokens (JWT-based, auto-rotating)
- All API calls use the installation token instead of the user token
- Rate limit: 5000 req/hr per installation (separate from user's limit)

## Implementation Notes
- Register a GitHub App with: contents:read, metadata:read, pull_requests:read permissions
- Add GITHUB_APP_ID, GITHUB_APP_PRIVATE_KEY, GITHUB_APP_INSTALLATION_ID to env
- Use @octokit/app to generate installation tokens
- Implement token rotation: cache token, refresh 5 min before expiry

## GSSoC Info
- Difficulty: Advanced — 55 pts
- Area: Infrastructure / Auth

Ridanshi:
## Summary

- Introduces `src/lib/github-app.ts` — a dedicated GitHub App service that generates RS256 JWTs, exchanges them for installation access tokens, and maintains an in-process cache with proactive refresh ~5 minutes before expiry.
- Concurrent callers share a single in-flight refresh request; token failures evict the cache so the next call retries cleanly.
- `src/lib/public-profile-data.ts` updated to call `resolveServerGitHubToken()` so public profile pages prefer the installation token (5 000 req/hr per installation) over the legacy `GITHUB_TOKEN` PAT.
- All three new env vars are fully optional — existing deployments are unaffected.

### Current state (before this PR)
- All GitHub API calls use the signed-in user's OAuth token from the NextAuth session.
- Server-side background code (`public-profile-data.ts`, weekly digest) falls back to `GITHUB_TOKEN` (a PAT) or unauthenticated requests.
- No GitHub App support existed.

### After this PR
- A GitHub App installation token can be configured via three env vars.
- When configured, it is preferred for server-side, non-user-scoped calls (public profiles, future background jobs).
- User-facing metric routes still use the per-user OAuth token — no change to the authenticated experience.
- Fallback chain: **App installation token → `GITHUB_TOKEN` PAT → unauthenticated**.

## Key exports (`src/lib/github-app.ts`)

| Export | Purpose |
|---|---|
| `isGitHubAppConfigured()` | Guard for optional App code paths |
| `readAppConfig()` | Validated env-var reader; normalises literal `\n` in PEM keys |
| `buildAppJwt(appId, privateKey)` | Signs an RS256 JWT for App-level auth |
| `fetchInstallationToken(config)` | `POST /app/installations/{id}/access_tokens` |
| `getInstallationToken()` | Cached, auto-refreshing token accessor |
| `resolveServerGitHubToken()` | Priority chain: App → PAT → `undefined` |
| `clearTokenCache()` | Cache eviction for error recovery and tests |
| `getInstallationRateLimitInfo()` | Rate-limit diagnostics for health endpoints |

## Token lifecycle

1. On first call to `getInstallationToken()`, a 10-minute RS256 JWT is signed with the App's private key.
2. The JWT is exchanged for a bearer token (GitHub issues these with a 1-hour TTL).
3. The token is stored in memory with its expiry timestamp.
4. Subsequent calls return the cached token until 5 minutes before expiry.
5. Near-expiry triggers a single refresh; concurrent callers await the same in-flight promise.
6. On network or HTTP failure the cache is cleared so the next call retries.

## Environment variables

All three must be set to enable the App path; any missing var leaves `isGitHubAppConfigured()` returning `false` and callers fall back to existing behaviour.

| Variable | Description |
|---|---|
| `GITHUB_APP_ID` | Numeric App ID (App settings → *About this App*) |
| `GITHUB_APP_PRIVATE_KEY` | PEM-encoded RSA private key. Literal `\n` escaping is normalised so the key can be stored as a single line in Vercel / Railway dashboards. |
| `GITHUB_APP_INSTALLATION_ID` | Installation ID — appears in the URL when viewing the App's installation page: `/settings/installations/{id}` |

## Required GitHub App permissions

- `contents:read`
- `metadata:read`
- `pull_requests:read`

## Files modified

| File | Change |
|---|---|
| `src/lib/github-app.ts` | **New** — complete service layer |
| `src/lib/public-profile-data.ts` | Use `resolveServerGitHubToken()` instead of `process.env.GITHUB_TOKEN` |
| `test/github-app.test.ts` | **New** — 35 tests |

## Security

- Private key is read from env vars and never logged; the JWT and installation tokens are also never written to logs.
- `buildAppJwt` back-dates `iat` by 60 seconds to absorb server clock skew.
- JWT lifetime is capped at 10 minutes (GitHub's maximum).
- Installation tokens expire in ~1 hour and are rotated automatically.

## Test results

```
✓ test/github-app.test.ts (35 tests) — 35 passed
✓ test/public-profile-data.test.ts (9 tests) — 9 passed
✓ test/public-profile-uuid-exposure.test.ts (8 tests) — 8 passed
```

Closes #238

### DIFF
--- a/src/lib/github-app.ts
+++ b/src/lib/github-app.ts
@@ -1,0 +1,358 @@
+/**
+ * GitHub App service layer.
+ *
+ * Provides installation access tokens using GitHub App credentials so
+ * DevTrack can make authenticated GitHub API calls from server-side code
+ * (public profile pages, background digests, achievement syncs) without
+ * relying on an individual user's OAuth token.
+ *
+ * How it works
+ * â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€
+ * 1. Sign a short-lived JWT (RS256, 10 min TTL) using the App's private key.
+ * 2. POST the JWT to /app/installations/{id}/access_tokens to obtain a
+ *    standard bearer token (GitHub issues these with a 1-hour lifetime).
+ * 3. Cache the bearer token in memory and refresh it ~5 minutes before it
+ *    expires so callers always receive a valid token.
+ * 4. Concurrent callers share a single in-flight refresh request â€” no
+ *    duplicate token generation on bursts.
+ *
+ * Environment variables
+ * â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€
+ *   GITHUB_APP_ID              Numeric App ID (App settings â†’ About this App)
+ *   GITHUB_APP_PRIVATE_KEY     PEM-encoded RSA private key.  Literal \n
+ *                              escaping (as stored in Vercel / Railway env
+ *                              dashboards) is normalised automatically.
+ *   GITHUB_APP_INSTALLATION_ID Installation ID for the target account/org.
+ *                              Find it in the App's installation URL:
+ *                              /settings/installations/{id}
+ *
+ * Required permissions on the GitHub App
+ * â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€
+ *   contents:read    â€” read repository file trees and commit data
+ *   metadata:read    â€” read public repository metadata
+ *   pull_requests:read â€” read PR lists and review state
+ *
+ * Backwards compatibility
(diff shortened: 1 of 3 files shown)


Priyanshu-byte-coder:
This PR has conflicts with main. Please rebase:
```
git fetch origin && git rebase origin/main
```
Resolve any conflicts, then push. We'll review and merge once CI passes.

PR #2264, opened 2026-06-10
fix: show at least 1% progress for fractional goal completion

### LINKED ISSUE #1895: [BUG] GoalTracker shows 0% progress for goals with fractional completion < 0.5%
## Describe the bug
The GoalTracker progress bar uses `Math.round()` to display progress percentage. For goals where actual / target < 0.005, this rounds down to **0%** even if there has been some activity, which is demotivating.

## Expected behaviour
Use `Math.ceil()` (or show one decimal place) for fractional progress so users can see any positive progress.

## Suggested fix
In `src/components/GoalTracker.tsx` (or wherever progress % is calculated):
```ts
// Before
const pct = Math.round((actual / target) * 100);
// After
const pct = actual > 0 ? Math.max(1, Math.round((actual / target) * 100)) : 0;
```

## Acceptance criteria
- [ ] Goals with any non-zero progress show at least 1%
- [ ] 0% is only shown when actual === 0

purvasatav:
## Summary
Fix GoalTracker showing 0% progress for goals with fractional completion below 0.5%.

Closes #1895

---
## Type of Change

---
## What Changed
- `src/components/GoalTracker.tsx` line 498: replaced `Math.round()` with a guard that returns minimum 1% for any real progress, and 0% if target is missing or zero
- `src/lib/goals/share.ts` line 6: same fix applied to the shared progress calculation utility

---
## How to Test
1. Create a goal with a large target (e.g. 500 commits)
2. Set current progress to a small value (e.g. 2)
3. Check the progress bar

**Expected result:** Progress bar shows 1% instead of 0%

---
## Checklist

### DIFF
--- a/src/components/GoalTracker.tsx
+++ b/src/components/GoalTracker.tsx
@@ -495,7 +495,7 @@ export default function GoalTracker() {
           {goals.map((goal) => {
             const pct =
               goal.current > 0
-                ? Math.max(1, Math.min(Math.round((goal.current / goal.target) * 100), 100))
+                ? (() => { if (!goal.target || goal.target <= 0) return 0; const raw = (goal.current / goal.target) * 100; if (raw <= 0) return 0; if (raw < 1) return 1; return Math.min(Math.round(raw), 100); })()
                 : 0;
             const isDeleting = deletingId === goal.id;
             const completed = goal.current >= goal.target;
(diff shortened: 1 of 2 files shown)


Priyanshu-byte-coder:
Closing — the existing code already handles the minimum 1% case correctly via `Math.max(1, Math.min(Math.round(...), 100))`. The replacement IIFE adds complexity without fixing an actual bug.

PR #2287, opened 2026-06-10
fix: prevent silent goal erasure when loadGoals fails 

### LINKED ISSUE #2285: [BUG] loadGoals silently replaces all visible goals with an empty list when the API returns an error response
### Bug Description

When `/api/goals` returns any error response (HTTP 401, 500, 503, etc.), all goals **disappear from the Goals widget** without any error message being shown. The widget reverts to the empty state: *"No goals yet. Track your coding progress by creating a goal."* — actively misleading the user into thinking their data is gone.

### Root cause

In [`src/components/GoalTracker.tsx`](../src/components/GoalTracker.tsx), the `loadGoals` function (lines 60–66) does not check `response.ok` before calling `.json()`:

```ts
const loadGoals = useCallback(async () => {
  const response = await fetch("/api/goals");
  const data: { goals: Goal[] } = await response.json(); // never throws on 4xx/5xx
  const fetchedGoals = data.goals ?? [];                  // undefined → []
  setGoals(fetchedGoals);                                  // wipes the UI
  return fetchedGoals;
}, []);
```

`fetch` only throws on network failures. On a 401 or 500, `response.json()` **resolves successfully** with the error body (e.g. `{ error: "Unauthorized" }`). Because this body has no `goals` field, `data.goals` is `undefined`, and `?? []` converts it to an empty array. `setGoals([])` is then called, clearing the UI.

Since no exception is thrown, none of the callers' `.catch()` blocks trigger, and no error state is set. The widget silently renders as empty.

`loadGoals` is called in **four places**, all of which are affected:

| Call site | Trigger |
|---|---|
| Line 105 | Initial component mount |
| Line 127 | `devtrack:sync` window event |
| Line 170 | After a new goal is created |
| Line 94 (via `handleSync`) | After clicking Refresh |

### Expected behaviour

When `/api/goals` returns a non-OK response, `loadGoals` should:
1. **Not** call `setGoals` with a new value (preserve the currently visible list).
2. **Throw** (or return an error signal) so callers can surface an error message to the user.

### Proposed fix

Add a `response.ok` guard and throw on error:

```ts
const loadGoals = useCallback(async () => {
  const response = await fetch("/api/goals");
  if (!response.ok) {
    throw new Error(`Failed to load goals (HTTP ${response.status})`);
  }
  const data: { goals: Goal[] } = await response.json();
  const fetchedGoals = data.goals ?? [];
  setGoals(fetchedGoals);
  return fetchedGoals;
}, []);
```

Then update the four call sites to catch the thrown error and call `setSyncError` (or a dedicated `setLoadError`) with a user-facing message instead of swallowing it silently.

### Files affected

- [`src/components/GoalTracker.tsx`](../src/components/GoalTracker.tsx) — `loadGoals` callback (lines 60–66) and its four call sites (lines 94, 105, 127, 170)

### Environment

- Affects all deployments; reproducible on any transient server error or session expiry.
- No environment-specific configuration required.

### Steps to reproduce

1. Create one or more goals on the dashboard.
2. Simulate a server error on `/api/goals` (e.g. intercept in browser DevTools → Network → right-click the request → Block request URL, then trigger a refresh).
3. Reload the page or click the Refresh button in the Goals widget.

**Expected:** Goals remain visible; an error message is shown (e.g. "Failed to load goals. Please try again.").  
**Actual:** All goals are replaced with an empty list. No error state is shown.

### Affected Area

Goal Tracker

just-tanvi:
## Description
This PR fixes a critical UX bug where the Goals widget would silently clear all visible goals and revert to the empty state if the `/api/goals` endpoint returned a non-OK response (e.g. `401 Unauthorized`, `500 Internal Server Error`).

Previously, `loadGoals` called `.json()` without verifying `response.ok`. Since error responses return an error payload (without a `goals` array), `data.goals` would be `undefined`. The widget defaulted this to an empty array and called `setGoals([])`, effectively wiping the user's view without giving them any feedback that a network/server issue had occurred.

## Changes Made
- Added a `!response.ok` guard in `loadGoals` (inside `src/components/GoalTracker.tsx`) that immediately throws an error containing the HTTP status code.
- Updated the `.catch(() => {})` blocks at the three affected call sites to explicitly catch these network errors and call `setSyncError(...)` with an appropriate, user-facing error message.
- Adjusted test assertions to ensure tests accurately reflect the widget's gracefully handled error states.

## Testing
- Verified that all unit tests for the `GoalTracker` component (`npx vitest run test/GoalTracker.test.ts`) pass successfully.
- Verified that a failing `loadGoals` call now preserves the current goals in the UI and correctly bubbles the error up to the UI banner.

closes #2285


just-tanvi:
@Priyanshu-byte-coder I will try to fix the 2 failing checks in the pr but it might take me some time, Thankyou for understanding !

PR #2305, opened 2026-06-10
fix(rooms): wire ?after= param to getRoomMessagesSince in messages GET route

### LINKED ISSUE #2301: [BUG] Room chat polling ignores ?after= parameter — always fetches full message history on every poll
## Bug Description

The `MessageFeed` component polls `/api/rooms/:roomId/messages?after=<timestamp>` every 5 seconds to fetch only messages newer than the last seen one. However, the API route handler reads `?before=` (not `?after=`) from the query string, so the `after` parameter is silently ignored on every poll.

As a result, the server always returns the **same last 50 messages** on every request instead of only the new messages since the last poll. The client-side deduplication (`existingIds.has(m.id)`) prevents visible duplicates, but it also means:
- The full 50-message payload is transferred over the network every 5 seconds per client, regardless of whether any new messages exist.
- In rooms with more than 50 messages, any messages that arrive between polls and fall outside the most-recent-50 window would be missed.
- `getRoomMessagesSince()` already exists in `src/lib/supabase-rooms.ts` (line 52) for exactly this use case but is never called by the GET route.

## Steps to Reproduce

1. Open a collaboration room that has more than a few messages.
2. Open the browser's Network tab.
3. Watch the polling requests to `/api/rooms/:roomId/messages?after=<timestamp>`.
4. Observe that every response returns the same set of messages (up to 50) regardless of the `after` timestamp sent.

## Affected Area

Collaboration Rooms

## Additional Context

**Root cause — parameter name mismatch:**

Client (`src/components/rooms/MessageFeed.tsx`, line 42):
```ts
const res = await fetch(
  `/api/rooms/${roomId}/messages?after=${encodeURIComponent(after)}`
);
```

Server (`src/app/api/rooms/[roomId]/messages/route.ts`, lines 18–19):
```ts
const before = url.searchParams.get('before') ?? undefined;  // reads 'before', not 'after'
const messages = await getRoomMessages(roomId, 50, before);   // 'before' is always undefined
```

`getRoomMessagesSince()` in `src/lib/supabase-rooms.ts` (line 52) already implements the correct `gt("created_at", after)` Supabase query but is never wired up in the GET route.

**Suggested fix** in `src/app/api/rooms/[roomId]/messages/route.ts`:
```ts
const after  = url.searchParams.get('after')  ?? undefined;
const before = url.searchParams.get('before') ?? undefined;

const messages = after
  ? await getRoomMessagesSince(roomId, after)   // incremental poll — only new messages
  : await getRoomMessages(roomId, 50, before);  // initial load / older-message pagination
```

**Environment:** Both (the route logic is the same in all environments)

nyxsky404:
## Summary

The room chat polling loop in `MessageFeed` sends `?after=<timestamp>` to fetch only messages newer than the last seen one. The GET handler was reading `?before=` exclusively, so `after` was silently ignored — every poll returned the same last-50 messages instead of only the new ones. This PR wires the `after` param to the already-existing `getRoomMessagesSince` helper.

Closes #2301

---

## Type of Change

---

## What Changed

- `src/app/api/rooms/[roomId]/messages/route.ts` — added `after` param read from query string; branched to `getRoomMessagesSince(roomId, after)` for incremental polls and kept `getRoomMessages(roomId, 50, before)` for initial loads and reverse pagination
- Added `getRoomMessagesSince` to the import from `@/lib/supabase-rooms`

---

## How to Test

1. Open a collaboration room that already has some messages.
2. Open the browser Network tab and filter by the room's message endpoint.
3. In another browser session (same room), send a new message.
4. Observe the polling request: the response should now contain **only the new message**, not the full history.
5. Confirm the new message appears in the first session within ~5 seconds.

**Expected result:** Each poll response returns only messages sent after the last-known timestamp. Initial page load still returns the most recent 50 messages as before.

---

## Checklist

- [x] No TypeScript errors (`pnpm run type-check` — errors shown are all pre-existing)
- [x] No new tests required — the logic change is a direct routing fix to an already-tested helper (`getRoomMessagesSince` has existing coverage in `test/rooms-messages.test.ts`)

### DIFF
--- a/src/app/api/rooms/[roomId]/messages/route.ts
+++ b/src/app/api/rooms/[roomId]/messages/route.ts
@@ -1,6 +1,6 @@
 import { getServerSession } from 'next-auth';
 import { authOptions } from '@/lib/auth';
-import { getRoomById, getRoomMessages, sendRoomMessage } from '@/lib/supabase-rooms';
+import { getRoomById, getRoomMessages, getRoomMessagesSince, sendRoomMessage } from '@/lib/supabase-rooms';
 import { validateTextInput } from '@/lib/sanitize';
 import { NextResponse } from 'next/server';
 
@@ -15,8 +15,11 @@ export async function GET(
   const room = await getRoomById(roomId, session.user.name);
   if (!room) return NextResponse.json({ error: 'Not found' }, { status: 404 });
   const url = new URL(req.url);
+  const after = url.searchParams.get('after') ?? undefined;
   const before = url.searchParams.get('before') ?? undefined;
-  const messages = await getRoomMessages(roomId, 50, before);
+  const messages = after
+    ? await getRoomMessagesSince(roomId, after)
+    : await getRoomMessages(roomId, 50, before);
   return NextResponse.json(messages);
 }
 


nyxsky404:
The CI failures on this PR (**Playwright smoke tests** and **Playwright visual regression**) are pre-existing broken tests on `main` — they are **not caused by the changes in this PR**.

**Root causes fixed:**
- `e2e/streak.spec.ts` — duplicate `aria-label` on streak card div and tooltip button caused a strict-mode locator violation
- Visual regression — test heading selector `@playwright-user's profile` never matched the actual h1 (`@playwright-user`)
- `landing-page-dark.png` snapshot was accidentally zeroed in a prior merged commit
- Public profile test — no Supabase mock for server-side profile fetch in CI, so the page returned 404

Fixes have been applied to this branch. CI is Green.

Priyanshu-byte-coder:
This PR has merge conflicts with the current `main` branch. Please rebase or merge `main` into your branch to resolve the conflicts, then push the updated branch. Once conflicts are resolved, this PR will be merged.

```bash
git fetch origin main
git merge origin/main
# resolve conflicts
git push
```

PR #2433, opened 2026-06-14
fix: prevent digest cooldown consumption when RESEND_API_KEY is missing

### LINKED ISSUE #2415: [BUG] Weekly digest cron treats missing RESEND_API_KEY as a successful send, consuming the 6-day cooldown with no email sent
## What happened?

In `src/app/api/cron/weekly-digest/route.ts` (lines 85–88), the `sendEmail()` function returns `{ ok: true }` when `RESEND_API_KEY` is absent:

```ts
const apiKey = process.env.RESEND_API_KEY;
if (!apiKey) {
  return { ok: true };  // ← treated as sent successfully
}
```

The outer batch loop treats this as a successful send and calls `recordDigestSent(userId)`, which sets `last_digest_sent_at` to the current timestamp. This starts the **6-day cooldown** for that user.

The result: self-hosted operators who deploy without configuring Resend will see the cron return `sentCount: N` with zero emails actually delivered. Every user's cooldown is consumed, and they will be silently skipped for the next 6 days on every subsequent cron run — indefinitely, as long as `RESEND_API_KEY` remains unset.

## What did you expect?

When `RESEND_API_KEY` is not configured, the cron should either:
1. Skip sending and **not** record a `last_digest_sent_at` (so the user is retried next week), or
2. 2. Return a clear error indicating email is unconfigured
The response should never report `sentCount > 0` when no emails were sent.

## Affected file

`src/app/api/cron/weekly-digest/route.ts` — lines 85–88 (`sendEmail` early return) and lines 126–132 (`recordDigestSent` call site)

## Suggested Fix

Change `sendEmail` to distinguish between "skipped" and "sent":

```ts
if (!apiKey) {
  return { ok: true, skipped: true };  // not an error, but not a real send
}
```

Then in the batch loop, only call `recordDigestSent()` when the email was actually sent:

```ts
const result = await sendEmail({ to, subject, html, text });
if (result.ok && !result.skipped) {
  await recordDigestSent(userId);
  sentCount++;
} else if (result.skipped) {
  skippedCount++;  // count separately, don't consume cooldown
} else {
  failedCount++;
}
```

Also add a warning to the cron response body when `RESEND_API_KEY` is absent, so operators are alerted.

### Steps to Reproduce

2. Click on '...'
3. See error

### Affected Area

Dashboard

### Screenshots

_No response_

### Browser & OS

_No response_

### Environment

None

### Additional Context

_No response_

Sricharan106:
## Summary

Fixes an issue where the weekly digest cron reported emails as successfully sent when `RESEND_API_KEY` was not configured. Previously, users were marked as having received a digest and entered the 6-day cooldown period despite no email being delivered.

Closes #2415

---

## Type of Change

* [x] 🐛 Bug fix (non-breaking change that fixes an issue)

---

## What Changed

* Updated `sendEmail()` to distinguish between successfully sent and skipped emails when `RESEND_API_KEY` is not configured.
* Prevented `last_digest_sent_at` from being updated for skipped emails.
* Added separate tracking for skipped emails so they do not consume the weekly digest cooldown period.
* Corrected cron response metrics to avoid reporting emails as sent when no delivery occurred.

---

## How to Test

1. Run the weekly digest cron with `RESEND_API_KEY` unset.
2. Verify the cron completes without sending emails.
3. Confirm users' `last_digest_sent_at` values are not updated and that skipped emails are reported separately.

**Expected result:**

* No emails are sent.
* Users are not placed into the cooldown window.
* `sentCount` remains accurate.
* Skipped deliveries are tracked separately.

---

## Additional Context

Before this change, deployments without a configured Resend API key could report successful email delivery while silently preventing users from receiving future digests due to the cooldown mechanism. This fix ensures cooldown timestamps are only recorded after a real email delivery attempt succeeds.


Sricharan106:
@Priyanshu-byte-coder can you review the pr when you got time

PR #2667, opened 2026-06-22
Refactor/api payload

Mohitmhatre32:
<html>
<body>
[...]

PR #2669, opened 2026-06-22
fix(webhooks): replace isSafeUrl+fetch with pinnedFetch in test and r…

### LINKED ISSUE #2564: [BUG] SSRF protection bypassable via DNS rebinding (TOCTOU) in custom webhook delivery
### Bug Description

`src/lib/webhooks.ts` validates a webhook URL using `isSafeUrl()` from `src/lib/ssrf-protection.ts` (which resolves the hostname via DNS and rejects it if it points to a private/internal IP), but then makes the **actual outgoing request with a separate `fetch(webhook.url, ...)` call**. Since DNS is resolved twice — once during validation, once during the real request — there's a time-of-check-to-time-of-use (TOCTOU) gap.
 
**Expected:** A webhook URL that resolves to a private/internal IP at request time should never be reachable, regardless of what it resolved to during validation.
 
**Actual:** An attacker can register a domain with a very short DNS TTL and configure it to return a public IP when `isSafeUrl()` checks it, then flip the DNS record to resolve to an internal IP (e.g. `169.254.169.254` cloud metadata, `10.x.x.x`, `127.0.0.1`) by the time the real `fetch()` call fires moments later. This lets an attacker use DevTrack's server as a proxy to probe/attack internal infrastructure (classic SSRF via DNS rebinding).

### Steps to Reproduce

1. Register/control a test domain (or use a local DNS server) where you can change the A record quickly.
2. Set up the domain to initially resolve to a public IP (e.g. `1.1.1.1`).
3. Create a custom webhook in DevTrack pointing to `http://attacker-controlled-domain.test/hook` via `POST /api/webhooks/custom`. This passes `isSafeUrl()` because at validation time DNS resolves to a public IP.
4. Immediately before/while the webhook delivery fires (e.g. trigger via `POST /api/webhooks/custom/[id]/test`), flip the DNS record for the domain to resolve to an internal address (e.g. `169.254.169.254` or `127.0.0.1`).
5. Observe that `fetch(webhook.url, ...)` in `src/lib/webhooks.ts` (around the delivery logic, ~line 100) re-resolves DNS and connects to the now-internal IP, since the safety check from step 3 is not re-validated against the IP actually used for the connection.
Relevant code path:
```js
// src/lib/webhooks.ts
const { isSafeUrl } = await import("./ssrf-protection");
const safe = await isSafeUrl(webhook.url);   // DNS resolved here, IP A
// ... later, separately ...
const response = await fetch(webhook.url, { ... });  // DNS resolved AGAIN here, can be IP B
```

### Affected Area

Dashboard

### Screenshots

N/A — this is a server-side network security logic bug, not a UI issue.

### Browser & OS

Not applicable — this is a backend/server-side vulnerability independent of browser or OS.

### Environment

None

### Additional Context

**Why this matters:** SSRF protections that validate a URL/hostname and then separately let the runtime (`fetch`) re-resolve DNS for the actual request are a well-known, exploitable pattern (CWE-918, DNS rebinding class). The validation becomes purely cosmetic against a motivated attacker with control over their own DNS.
 
**Suggested fix:** Resolve the hostname to an IP once, validate that IP with `isPrivateIP()`, and then perform the actual request against that **pinned IP** (e.g. using `http.Agent`/`https.Agent` with a custom `lookup` function, or setting the `Host` header and connecting directly to the validated IP) rather than letting `fetch()` re-resolve the original hostname. Should also add a redirect check, since `fetch` follows redirects by default — a validated URL could redirect (3xx) to an internal address, which would bypass `isSafeUrl()` entirely if redirects aren't disabled or re-validated per-hop.
 
**Severity:** High — SSRF can be used to access cloud metadata endpoints (potential credential theft), scan/attack internal network services, or hit other internal-only APIs that the server can reach but the public internet cannot.

Priyanshi-untitled:
## Summary
Fixes a DNS-rebinding TOCTOU SSRF vulnerability in custom webhook delivery where `isSafeUrl()` validated the URL but `fetch()` re-resolved DNS separately, allowing an attacker to flip a short-TTL DNS record between the check and the actual request to point at an internal IP. Replaces the validate-then-fetch pattern with a single `pinnedFetch()` helper that resolves DNS once, validates the IP, and connects directly to that pinned IP.

Closes #2564 

---

## Type of Change
- [ ] 📝 Documentation update
- [x] 🔒 Security fix
- [ ] 🧪 Tests only

---

## What Changed

- **`src/lib/ssrf-protection.ts`** — Added new `resolveAndValidateUrl()` function that resolves DNS exactly once, validates every returned IP against the private-range blocklist, and returns the pinned IP for use in the actual request. Marked existing `isSafeUrl()` as `@deprecated` for outbound-request use cases (still kept — used by `repo-analytics` route which is a separate concern).
- **`src/lib/pinned-fetch.ts`** *(new file)* — Added `pinnedFetch()` helper that calls `resolveAndValidateUrl()` internally and makes the HTTP(S) request directly to the pinned IP using Node's `http`/`https` modules (with correct `Host` header and TLS `servername` for SNI). Does not follow redirects automatically, preventing a redirect-based SSRF bypass.
- **`src/lib/webhooks.ts`** — Replaced `isSafeUrl() + fetch()` two-step with a single `pinnedFetch()` call in `dispatchWebhook()`. Removed the now-redundant `try/catch` wrapper since `pinnedFetch` resolves rather than throws on failure.
- **`src/app/api/webhooks/custom/[id]/test/route.ts`** — Replaced `isSafeUrl` import and separate `fetch()` call (lines 52–94) with `pinnedFetch()`. Removed `try/catch` block, now reads result directly from `pinnedFetch` return value.
- **`src/app/api/webhooks/custom/[id]/deliveries/[deliveryId]/retry/route.ts`** — Removed the redundant `isSafeUrl()` pre-check (lines 63–75) since `dispatchWebhook()` now handles validation internally via `pinnedFetch`. No other changes needed in this file.
- **`src/lib/pinned-fetch.test.ts`** *(new file)* — Unit tests for `resolveAndValidateUrl()` covering private IPv4 literals, loopback, cloud metadata address (`169.254.169.254`), public IP acceptance, and unsupported protocol rejection.

---

## How to Test

1. Run the test suite to verify the new unit tests pass:
   ```bash
   pnpm run test
   ```
2. Start the dev server and create a custom webhook pointing to a real public URL (e.g. `https://webhook.site/your-unique-id`):
   ```
   POST /api/webhooks/custom
   ```
3. Trigger a test delivery via the UI or:
   ```
   POST /api/webhooks/custom/[id]/test
   ```
4. Verify the delivery succeeds and appears in the webhook deliveries log.
5. Now try creating a webhook pointing to a private IP directly (e.g. `http://127.0.0.1/hook` or `http://10.0.0.1/hook`) and trigger delivery — verify it is blocked with a `400` error and the delivery is logged as failed with an SSRF error message.
6. Run lint and type-check:
   ```bash
   pnpm run lint
   pnpm run type-check
   pnpm run build
   ```

**Expected result:** Legitimate public webhook URLs deliver successfully. Any URL resolving to a private/internal IP — whether a literal IP or via DNS — is blocked at the `pinnedFetch` validation step before any network connection is made.

---

## Screenshots / Recordings

N/A — this is a backend security fix with no UI changes.

---

## Checklist
- [x] Added or updated tests where applicable (`src/lib/pinned-fetch.test.ts`)
- [x] Updated documentation / comments if behavior changed (`@deprecated` added to `isSafeUrl`)

---

N/A — no UI changes.

---

## Additional Context

**Why not just call `isSafeUrl()` twice (before and after fetch)?**
Still racy — there's always a window between the second check and the actual TCP connection. The only safe fix is to pin the IP at the OS/socket level so DNS is never re-queried.

**Why Node's `http`/`https` instead of `fetch` with a custom agent?**
Node's global `fetch` (undici under the hood) doesn't expose a stable `lookup`-override API in the versions this project targets. Using `http.request`/`https.request` with `hostname` set to the pinned IP gives complete control over the connection target without adding a new dependency.

**Why no redirect following in `pinnedFetch`?**
Each redirect hop would re-introduce the same TOCTOU gap — a validated URL redirecting to `http://169.254.169.254` would bypass all checks if redirects were followed blindly. Redirects are blocked entirely for now; supporting them safely would require re-validating each `Location` header through `resolveAndValidateUrl` and is tracked as a follow-up.

**`isSafeUrl` in `repo-analytics` route** — that route uses `isSafeUrl` for a read-only fetch of external repo data, which is lower risk but has the same underlying gap. Left as a separate follow-up issue to keep this PR focused.


Legit-Ox (changes requested):
## Review: SSRF DNS-rebinding fix

Core approach is correct and well-reasoned. Two security issues must be fixed before merge — see inline comments.

**Security blockers:**
1. `isPrivateIP()` is IPv4-only — IPv6 private ranges (`fc00::/7`, `fe80::/10`, ULA) bypass validation entirely
2. Raw IPv6 address passed to `http.request` hostname without brackets — breaks legitimate IPv6 connections

**Other issues:**
- `pnpm-workspace.yaml` has placeholder strings (same issue as PR #2667) — delete it
- Hindi/Hinglish comments in `webhooks.ts` — English-only codebase
- Broken indentation in `test/route.ts` new block (1-space vs 2-space)
- No response body size cap in `pinnedFetch` — OOM risk on malicious large responses
- `pinned-fetch.test.ts` only tests `resolveAndValidateUrl`, not `pinnedFetch` HTTP logic
- Missing newline at EOF in both new files

src/lib/ssrf-protection.ts
```diff
@@ -109,3 +114,54 @@ export function validateUrlBasic(url: string): boolean {
...
+    ? hostname.slice(1, -1) : hostname;
+
+  if (hostname === "localhost" || ipToCheck === "0.0.0.0" || ipToCheck === "::1") {
+    return { safe: false, reason: "Blocked hostname" };
+  }
+
+  // Literal IP — validate directly, no DNS needed
+  if (net.isIP(ipToCheck)) {
+    return isPrivateIP(ipToCheck)
+      ? { safe: false, reason: "Private IP literal" }
```
Legit-Ox, line 149: **Security: `isPrivateIP()` is IPv4-only — IPv6 private ranges bypass validation.**

`dns.lookup({ all: true })` can return IPv6 addresses. `isPrivateIP` converts to an IPv4 number — IPv6 addresses produce `NaN`, fail all range checks, and are marked safe.

Affected ranges not blocked: `fc00::/7` (ULA), `fe80::/10` (link-local), `fd00::/8`, `::ffff:0:0/96` (IPv4-mapped), etc.

Fix: check `net.isIPv6(addr)` before calling `isPrivateIP` and block private IPv6 ranges explicitly:
```ts
for (const addr of addresses) {
  if (net.isIPv6(addr) ? isPrivateIPv6(addr) : isPrivateIP(addr)) {
    return { safe: false, reason: "Resolved to private address" };
  }
}
```

src/lib/pinned-fetch.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (!v.safe || !v.pinnedIp || !v.hostname || !v.protocol) {
+    return { ok: false, error: v.reason ?? "URL failed SSRF validation" };
+  }
+
+  const parsed = new URL(url);
+  const transport = v.protocol === "https:" ? https : http;
+
+  return new Promise((resolve) => {
+    const req = transport.request(
+      {
```
Legit-Ox, line 38: **IPv6 addresses need bracket wrapping for `http.request` hostname.**

If `v.pinnedIp` is an IPv6 address (e.g. `2001:db8::1`), passing it raw to `hostname` causes a parse error. Node's `http.request` requires `[2001:db8::1]`.

Fix:
```ts
hostname: net.isIPv6(v.pinnedIp) ? `[${v.pinnedIp}]` : v.pinnedIp,
```

src/lib/webhooks.ts
```diff
@@ -1,6 +1,7 @@
 import { createHmac, randomBytes } from "crypto";
 import { supabaseAdmin } from "./supabase";
 import { encryptToken, decryptToken } from "./crypto";
+import { pinnedFetch } from "./pinned-fetch";  // ← naya import
```
Legit-Ox, line 5: Comment is in Hindi/Hinglish — project uses English throughout. Remove the `// ← naya import` suffix.

src/lib/webhooks.ts
```diff
@@ -79,79 +80,45 @@ export async function dispatchWebhook(
   const payloadString = JSON.stringify(payload);
   const signature = signPayload(payloadString, secret);
 
-  const { isSafeUrl } = await import("./ssrf-protection");
-  const safe = await isSafeUrl(webhook.url);
-  if (!safe) {
-    const errorMessage = "SSRF protection: blocked request to private/internal address";
+  // ✅ FIX: pinnedFetch ek hi baar DNS resolve karta hai aur us pinned IP
+  // pe directly connect karta hai — TOCTOU gap khatam
```
Legit-Ox, line 84: Comments in Hindi — replace with English or remove entirely:
```ts
// pinnedFetch resolves DNS once and connects to the pinned IP — eliminates TOCTOU gap
```

src/lib/pinned-fetch.ts
```diff
@@ -0,0 +1,63 @@
...
+        method: options.method,
+        headers: { ...options.headers, Host: v.hostname },
+        timeout: options.timeoutMs ?? 10_000,
+      },
+      (res) => {
+        const chunks: Buffer[] = [];
+        res.on("data", (c) => chunks.push(c));
+        res.on("end", () => resolve({
+          ok: (res.statusCode ?? 0) >= 200 && (res.statusCode ?? 0) < 300,
+          status: res.statusCode,
```
Legit-Ox, line 53: No response body size limit — a malicious endpoint returning a multi-GB response buffers entirely in memory. Add a cap:
```ts
const MAX_BODY = 1 * 1024 * 1024; // 1 MB
let size = 0;
res.on("data", (c) => {
  size += c.length;
  if (size > MAX_BODY) { req.destroy(); return; }
  chunks.push(c);
});
```

Legit-Ox (commented):
Delete pnpm-workspace.yaml.

Priyanshu-byte-coder:
Good SSRF fix (pinnedFetch > isSafeUrl + fetch). Two issues: (1) broken `pnpm-workspace.yaml` with placeholders — remove it, (2) webhook test route has broken indentation (1-space instead of 2). Please fix and rebase on main.

Priyanshu-byte-coder:
This is the right approach to the TOCTOU/DNS-rebinding gap — resolving once and connecting to the pinned IP via `https.request` with `servername` (SNI) + `Host` header keeps TLS valid, which the `fetch(hostname=ip)` alternative in #2669's sibling PR does not. Preferring this one. A few things needed before it can merge:

1. **Remove the non-English / emoji inline comments** (e.g. `// ← naya import`, `// ✅ FIX: … khatam`, `// SSRF block ya network failure …`). Codebase comments are English-only; please rewrite them plainly.
2. **Fix the indentation** in `webhooks/custom/[id]/test/route.ts` — the reformatted `testPayload`/`pinnedFetch` block is single-space indented and misaligned with the rest of the function.
3. **Update `test/webhooks.test.ts`** — `dispatchWebhook` no longer calls `isSafeUrl`, so the existing `isSafeUrl` mocks no longer gate behavior. Mock `pinnedFetch` (or `resolveAndValidateUrl`) instead, otherwise the SSRF-block test exercises the real network path.
4. Add trailing newlines to the changed files and drop the unrelated `pnpm-workspace.yaml` blank-line churn.

Get `pnpm test`, `pnpm lint`, `pnpm type-check` green locally and I'll merge.

PR #2671, opened 2026-06-22
fix(sanitize):decode HTML entities before stripping tags to prevent X…

### LINKED ISSUE #2565: [BUG] XSS bypass in stripHtml() due to decode-after-strip ordering (src/lib/sanitize.ts)
### Bug Description

`stripHtml()` in `src/lib/sanitize.ts` strips HTML tags **first** using a regex (`/<[^>]*>/g`), and only **after that** decodes HTML entities like `&lt;` and `&gt;` back into `<` and `>`. This means an attacker can submit an HTML-entity-encoded payload (which contains no raw `<`/`>` for the tag-stripping regex to catch), and the entity-decoding step that runs afterward will reconstruct the exact tag that was supposed to be removed.
 
**Expected:** Any HTML/script tags, encoded or not, should be neutralized by `stripHtml()`, since it's used by `validateTextInput()` for sanitizing user-supplied text fields.
 
**Actual:** Entity-encoded payloads like `&lt;script&gt;alert(1)&lt;/script&gt;` pass through `stripHtml()` and come out as a live `<script>alert(1)</script>` tag, fully defeating the sanitizer.

### Steps to Reproduce

1. Open a Node REPL or add a quick test file in the project.
2. Import/copy `stripHtml` from `src/lib/sanitize.ts`.
3. Run:
```js
   stripHtml("&lt;script&gt;alert(1)&lt;/script&gt;")
```
4. **Observe output:** `"<script>alert(1)</script>"` — a fully reconstituted, executable script tag.
5. Compare with a non-encoded payload:
```js
   stripHtml("<img src=x onerror=alert(1)>")
```
   which correctly returns `""`, showing the bypass is specific to the entity-encoded path.
6. Any API route that accepts free text and runs it through `validateTextInput()` (which calls `stripHtml()` internally) is exposed to this — e.g. profile bio fields, daily notes, goal descriptions, milestone titles, etc.

### Affected Area

Dashboard

### Screenshots

N/A — this is a logic bug reproducible via unit test/REPL, not a visual issue.

### Browser & OS

_No response_

### Environment

None

### Additional Context

**Root cause:** the two `.replace()` calls in `stripHtml()` run in the wrong order:
```js
export function stripHtml(value: string): string {
  return value
    .normalize("NFKC")
    .replace(/<[^>]*>/g, "")          // (1) strips raw tags
    .replace(/&(?:lt|gt|amp|quot|#x27|#39);/gi, (m) => { ... })  // (2) decodes entities AFTER stripping
    .trim();
}
```
Step (2) can re-introduce `<` and `>` characters into the string after step (1) has already finished looking for them, so a tag smuggled in as entities survives untouched.
 
**Suggested fix:** Either decode entities first and then strip tags, and run the tag-strip again after decoding (loop until stable, or just strip-only and never decode), or simplest: drop the entity-decoding step entirely from a "strip" function — a sanitizer shouldn't be re-introducing characters it just removed. Should also add a regression test asserting `stripHtml("&lt;script&gt;alert(1)&lt;/script&gt;")` contains no `<`/`>` characters.
 
**Severity:** High — this is a stored/reflected XSS vector wherever `validateTextInput`/`stripHtml` is the only sanitization layer before data is rendered.

Priyanshi-untitled:
## Summary
Fixes an XSS bypass in `stripHtml()` where entity-encoded payloads like `&lt;script&gt;alert(1)&lt;/script&gt;` survived sanitization and were reconstructed into live `<script>` tags. The root cause was the two `.replace()` calls running in the wrong order — entities were decoded *after* tags were stripped, so entity-smuggled tags were never seen by the tag-stripping regex. Fix: swap the order (decode first, strip second) and add regression tests.

Closes #2565 

---

## Type of Change
- [ ] 📝 Documentation update
- [x] 🔒 Security fix
- [ ] 🧪 Tests only

---

## What Changed

- **`src/lib/sanitize.ts`** — Swapped the order of the two `.replace()` calls inside `stripHtml()`. Entity decoding (`&lt;` → `<` etc.) now runs *before* tag stripping (`/<[^>]*>/g`), so tags smuggled in via entity encoding are decoded into raw tags first and then caught by the strip regex. Added an explanatory comment documenting why the order matters.
- **`src/lib/sanitize.test.ts`** *(new file)* — Regression tests for `stripHtml()` and `validateTextInput()` covering: raw tag stripping, entity-encoded tag bypass (the XSS vector), mixed encoded payloads, plain text passthrough, safe entity decoding, and `validateTextInput` edge cases (non-string input, empty-after-strip, over-maxLen).

---

## How to Test

1. Run the test suite — the new regression test directly asserts the bypass is closed:
   ```bash
   pnpm run test
   ```
2. Manual REPL verification:
   ```bash
   node -e "
   // paste the fixed stripHtml function here and run:
   console.log(stripHtml('&lt;script&gt;alert(1)&lt;/script&gt;'));
   // expected: ''  (empty string, not a live script tag)
   "
   ```
3. Run lint and type-check:
   ```bash
   pnpm run lint
   pnpm run type-check
   pnpm run build
   ```

**Expected result:**
- `stripHtml("&lt;script&gt;alert(1)&lt;/script&gt;")` returns `""` (was returning `"<script>alert(1)</script>"` before the fix)
- `stripHtml("")` still returns `""` (no regression)
- `stripHtml("hello &amp; world")` returns `"hello & world"` (safe entities still decoded correctly)

---

## Screenshots / Recordings

N/A — logic bug with no UI changes. Reproducible via unit test / REPL as shown above.

---

## Checklist
- [x] Added regression tests (`src/lib/sanitize.test.ts`)
- [x] Added explanatory comment in `stripHtml()` documenting why order matters

---

N/A — no UI changes.

---

## Additional Context

**Why not just remove the entity-decode step entirely?**
`validateTextInput()` is also used for display — plain text fields like bio or goal titles legitimately contain `&amp;`, `&quot;` etc. from user input that should render as `&`, `"`. Removing decode entirely would show raw entities to users. The correct fix is to decode first so the strip step can see and remove any tags that were hiding behind entities, which is what this PR does.

**Why not use a library like `dompurify`?**
DOMPurify requires a DOM environment (browser or jsdom) and adds a dependency. The existing function is close to correct — it just needed a one-line order swap. Keeping it lightweight and dependency-free is preferable for a server-side utility; the fix is minimal and the behavior is now fully covered by regression tests.

**Affected surfaces (all go through `validateTextInput` → `stripHtml`):**
profile bio, daily notes, goal titles/descriptions, milestone titles — any free-text field accepted by the API. All are now protected by this fix without any other changes needed.


Priyanshu-byte-coder:
Good XSS fix — entity decode before tag strip is the right order. But PR includes a broken `pnpm-workspace.yaml` with placeholder values that would overwrite existing config. Please remove that file and rebase on main.

Priyanshi-untitled:
@Priyanshu-byte-coder , I have fixed the issue !!

Priyanshu-byte-coder:
The decode-then-strip fix is already on main (src/lib/sanitize.ts decodes entities before stripping). What's left in this diff is only whitespace changes and tests duplicating test/sanitize.test.ts — and the repo convention keeps tests in test/, not src/lib/. Nothing left to merge here, but the analysis was right.

Priyanshu-byte-coder:
Tests look good and they import the real `stripHtml`/`validateTextInput`. Two things to fix first: (1) the `src/lib/sanitize.ts` change de-indents `let text = value.normalize("NFKC")` to column 0 inside the function — that's inconsistent and prettier/lint will reject it; please revert that line so the diff is tests-only. (2) Drop the unrelated `pnpm-workspace.yaml` blank-line removals. Also add a trailing newline to the new test file. Once it's a clean tests-only PR with `pnpm test` + `pnpm lint` green, happy to merge.

PR #2672, opened 2026-06-22
fix(rate-limit): only trust forwarding headers from configured truste…

### LINKED ISSUE #2566: [FEAT] Harden rate limiting with a trusted-proxy IP allowlist (stop trusting spoofable headers)
### Problem Statement

`getClientIp()` in `src/lib/rate-limit.ts` derives the client's IP directly from `cf-connecting-ip`, `x-real-ip`, and `x-forwarded-for` headers, in that order, with no verification that the request actually passed through a trusted proxy (e.g. Cloudflare or Vercel's edge). Since these headers are just regular HTTP headers, any client can set them to arbitrary values on a direct request.
 
This means anyone can get a fresh rate-limit bucket on every single request simply by sending a different `x-forwarded-for` value each time, completely defeating rate limiting on endpoints that rely on `getClientIp()` — including `auth-rate-limit.ts`, which is meant to slow down brute-force/abuse on authentication-adjacent flows. The rate limiter currently provides a false sense of protection rather than real protection.

### Proposed Solution

Make `getClientIp()` (or a new wrapper around it) aware of which deployment environment it's running in and only trust forwarding headers when the request is verified to have come from a known, trusted proxy:
 
1. Add a config flag / env var (e.g. `TRUSTED_PROXY = "vercel" | "cloudflare" | "none"`) so the behavior is explicit per deployment.
2. When `TRUSTED_PROXY = "vercel"`, rely on Vercel's own guaranteed header (Vercel sets `x-real-ip`/`x-forwarded-for` itself at the edge and it cannot be overridden by the client in their infra) and document that this is safe specifically because Vercel terminates the connection — but only honor it when running on Vercel (e.g. detect via `process.env.VERCEL`).
3. When `TRUSTED_PROXY = "cloudflare"`, only trust `cf-connecting-ip` if the request additionally includes a valid Cloudflare-only signal (e.g. validate against Cloudflare's published IP ranges for the immediate connecting peer, if accessible in the runtime, or require `CF-Worker`/a shared secret header set by a Cloudflare Worker in front of the app).
4. When no trusted proxy is configured (`"none"`), fall back to a non-spoofable identifier instead of trusting any forwarding header — e.g. combine a session-bound identifier (signed cookie/token) with a coarse fallback, rather than silently trusting `x-forwarded-for`.
5. Log/flag (in non-prod) when a forwarding header is present but not trusted, to help catch misconfigurations during rollout.
6. Update `auth-rate-limit.ts` and any other consumer to use the hardened resolver instead of calling `getClientIp()` directly.

### Feature Area

Dashboard

### Alternatives Considered

- **Switch entirely to Upstash Redis-based rate limiting keyed by authenticated user ID instead of IP** — solves spoofing for logged-in flows, but doesn't help pre-auth endpoints (e.g. signup/login attempts) where there's no user ID yet and IP is the only signal available.
- **Just trust `x-forwarded-for` unconditionally (status quo)** — simplest, but leaves rate limiting trivially bypassable, which defeats the purpose of having it on auth-sensitive routes.
- **Use a CAPTCHA/challenge instead of IP-based limiting** — adds friction for all users and is a much bigger UX change than fixing the IP resolution logic; could be a complementary future enhancement but doesn't replace the need for correct IP attribution.

### Acceptance Criteria

- [ ] `getClientIp()` (or its replacement) does not trust forwarding headers unless the deployment is explicitly configured to trust the proxy that set them.
- [ ] Sending a spoofed `x-forwarded-for`/`x-real-ip`/`cf-connecting-ip` header from a direct, untrusted client request does not bypass or reset the rate limit bucket.
- [ ] Behavior is documented per supported deployment target (Vercel, Cloudflare-fronted, self-hosted/none).
- [ ] `auth-rate-limit.ts` and other existing consumers continue to work correctly on the project's actual production deployment (Vercel) without regressing legitimate users sharing an IP (e.g. office NAT) into one bucket unfairly.
- [ ] Unit tests added covering: trusted-proxy header accepted, untrusted/spoofed header rejected/ignored, and the "no trusted proxy configured" fallback path.

### Additional Context

This pairs naturally with the existing `createMemoryFixedWindowRateLimiter` — the in-memory limiter is also per-instance on serverless (a separate, already-tracked concern), but fixing IP attribution is a prerequisite either way, since even a Redis-backed global limiter is useless if the key (client IP) can be freely chosen by the caller. Related reference: OWASP guidance on IP spoofing via `X-Forwarded-For` and trusted-proxy configuration (https://owasp.org/www-community/attacks/HTTP_Request_Smuggling and general reverse-proxy header trust guidance).

Priyanshi-untitled:
## Summary
Fixes a rate-limit IP spoofing vulnerability where `getClientIp()` unconditionally trusted `cf-connecting-ip`, `x-real-ip`, and `x-forwarded-for` headers regardless of whether the request actually came through a trusted proxy. Any client could send a different spoofed IP header per request to get a fresh rate-limit bucket, completely defeating IP-based limiting on auth-sensitive endpoints. Adds a `TRUSTED_PROXY` env var so headers are only trusted when the proxy that sets them is explicitly configured.

Closes #2566 

---

## Type of Change
- [ ] 📝 Documentation update
- [x] 🔒 Security fix
- [ ] 🧪 Tests only

---

## What Changed

- **`src/lib/rate-limit.ts`** — Added `getTrustedProxy()` which reads a new `TRUSTED_PROXY` env var (`"vercel" | "cloudflare" | "none"`) and auto-detects Vercel via `process.env.VERCEL === "1"`. Updated `getClientIp()` to only read forwarding headers when the relevant proxy is trusted: `x-real-ip` / `x-forwarded-for` for Vercel, `cf-connecting-ip` for Cloudflare, and `"unknown"` fallback when no trusted proxy is set (headers ignored entirely). Added a dev-mode `console.warn` when forwarding headers are present but `TRUSTED_PROXY` is not configured, to catch deployment misconfigurations early.
- **`.env.example`** — Added `TRUSTED_PROXY=vercel` with explanatory comment documenting all valid values and when to use each.
- **`src/lib/rate-limit.test.ts`** — Added four new tests covering: `TRUSTED_PROXY=vercel` trusts `x-real-ip` and ignores `cf-connecting-ip`, `TRUSTED_PROXY=cloudflare` trusts `cf-connecting-ip`, `TRUSTED_PROXY=none` returns `"unknown"` and ignores all spoofed forwarding headers, and auto-detection via `VERCEL=1` when `TRUSTED_PROXY` is not set.

---

## How to Test

1. Run the test suite — new tests directly assert the spoofing-prevention behavior:
   ```bash
   pnpm run test
   ```
2. To manually verify on Vercel: ensure `TRUSTED_PROXY=vercel` is set in Vercel environment variables (or leave unset — `VERCEL=1` is set automatically by Vercel and triggers auto-detection).
3. To verify spoofing is blocked locally:
   ```bash
   TRUSTED_PROXY=none pnpm run dev
   ```
   Then send a request with a spoofed header:
   ```bash
   curl -H "x-forwarded-for: 1.2.3.4" http://localhost:3000/api/public/someuser
   ```
   Rate limiting will key on `"unknown"` instead of the spoofed IP, confirming headers are ignored.
4. Run lint, type-check, build:
   ```bash
   pnpm run lint && pnpm run type-check && pnpm run build
   ```

**Expected result:** With `TRUSTED_PROXY=none`, all forwarding headers are ignored and rate limiting keys on `"unknown"`. With `TRUSTED_PROXY=vercel`, only `x-real-ip` / `x-forwarded-for` are read (set by Vercel's edge, not spoofable by the client). With `TRUSTED_PROXY=cloudflare`, only `cf-connecting-ip` is trusted.

---

## Screenshots / Recordings

N/A — backend security fix with no UI changes.

---

## Checklist
- [x] Added tests for all three proxy modes + auto-detection
- [x] Updated `.env.example` with documentation for the new env var

---

N/A — no UI changes.

---

## Additional Context

**Why `"unknown"` instead of blocking the request when no proxy is configured?**
Blocking all requests when `TRUSTED_PROXY` is not set would break self-hosted deployments that have no reverse proxy. Keying rate limits on `"unknown"` means all such clients share one bucket — not ideal but still a meaningful throttle, and importantly it cannot be bypassed by spoofing headers.

**Does this affect `auth-rate-limit.ts`?**
Yes — `auth-rate-limit.ts` calls `checkAuthRateLimit(ip)` where `ip` comes from `getClientIp()` (in `middleware.ts`). With this fix, the IP passed to `checkAuthRateLimit` is now the actual verified client IP rather than an attacker-controlled value, so auth brute-force protection now actually works.

**Does this affect `middleware.ts`'s own `getIp()`?**
`middleware.ts` has its own local `getIp()` function (separate from `getClientIp()` in `rate-limit.ts`) with the same spoofing vulnerability. That is a separate but related issue and should be fixed as a follow-up to keep this PR focused.

**Existing behavior on Vercel (current production):**
`VERCEL=1` is set automatically by Vercel, so auto-detection kicks in and behavior is identical to before for the Vercel deployment — no regression for current users.


Priyanshu-byte-coder:
Good security fix for IP spoofing via forwarding headers. However, the PR includes a broken `pnpm-workspace.yaml` with placeholder values ("set this to true or false") that would overwrite the existing config and break the build. Please remove that file from your branch and rebase on main.

Priyanshi-untitled:
@Priyanshu-byte-coder , Fixed! Removed the broken pnpm-workspace.yaml and restored the original from main. Thanks for catching it!

PR #2702, opened 2026-06-22
test : added unit tests for github-rate-limit utilities

### LINKED ISSUE #2697: test : add unit tests for github-rate-limit utilities
Summary of What Needs to be Done: Add unit tests for the utility functions in src/lib/github-rate-limit.ts: getGitHubRateLimitDetails, throwIfGitHubRateLimited, and githubRateLimitResponse.

Changes that Needs to be Made: Create test/github-rate-limit.test.ts. Mock Response objects (with .status and .headers.get) to cover: 403/429 with remaining=0 and valid reset header, 403/429 with remaining=0 and missing reset header, 403/429 with remaining>0 (should return null), 200 OK (should return null), and NaN reset header. Also test that GitHubRateLimitError is correctly thrown and that githubRateLimitResponse produces the expected JSON shape.

Impact that it would Provide: This module is the first line of defense against rate limit exhaustion cascading through the app. Proper test coverage ensures the detection logic is correct and the error response shape stays stable across refactors.

Hello @Priyanshu-byte-coder, please assign this issue to me (@tmdeveloper007) so I can open a focused PR.

tmdeveloper007:
Closes #2697.

Summary of What Has Been Done: Added unit tests for getGitHubRateLimitDetails, throwIfGitHubRateLimited, and githubRateLimitResponse from src/lib/github-rate-limit.ts. All three functions were already exported from the source module.

Changes Made:
- Created test/github-rate-limit.test.ts with 16 test cases covering: status code + remaining header combinations, missing and NaN reset headers, GitHubRateLimitError throwing and its details, githubRateLimitResponse null handling for non-error inputs, and the 429 JSON response shape with rate limit metadata

Impact it Made: github-rate-limit is the gatekeeper for rate limit errors across GitHub API calls. These tests ensure the detection logic is correct and the error response shape stays stable, preventing cascading failures from undetected rate limiting.

### DIFF
--- a/test/github-rate-limit.test.ts
+++ b/test/github-rate-limit.test.ts
@@ -1,4 +1,4 @@
-﻿import { describe, it, expect } from "vitest";
+import { describe, expect, it } from "vitest";
 import {
   getGitHubRateLimitDetails,
   throwIfGitHubRateLimited,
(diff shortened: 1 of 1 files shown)


Priyanshu-byte-coder:
Good rewrite — using real Response/Headers objects instead of the mock helper is strictly better. One thing to restore before merge:

The old file had a test for the secondary rate-limit path (403 with retry-after header). That case was removed and is not replaced anywhere. If that code path exists in the source, please add it back.

Otherwise CI is clean and the rest of the coverage improvement is solid.

Issue #2695, opened 2026-06-22
test : add unit tests for achievement-estimators helper functions


tmdeveloper007:
Summary of What Needs to be Done: Add unit tests for the pure helper functions in src/lib/achievement-estimators.ts: calculateNextTier and calculatePercentage. Both functions are synchronous and have no external dependencies, making them ideal test targets.

Changes that Need to be Made: Create test/achievement-estimators.test.ts (matching vitest.config.ts include pattern). Cover: calculateNextTier with current below first tier, current equal to a tier, current above all tiers (maxed), and empty tiers array. Cover calculatePercentage with current below nextTier, current at nextTier, current above nextTier, and nextTier=null (maxed).

Impact that it would Provide: Improves test coverage on a module that currently has no tests. These are foundational functions used across multiple achievement progress calculations, so regression protection here has high value.

Hello @Priyanshu-byte-coder, please assign this issue to me (@tmdeveloper007) so I can open a focused PR.

desireddymohithreddy0925:
Greetings @Umbrella-io Team,

I would like to work on this issue under GSSoC 2026.

I understand that we need to add unit tests for the helper functions in src/lib/achievement-estimators.ts, specifically calculateNextTier and calculatePercentage. These functions are pure, synchronous, and have no external dependencies, making them suitable for comprehensive unit testing.

I plan to work on the following:

1. Creating a new test file: test/achievement-estimators.test.ts (aligned with vitest.config.ts include pattern).
2. Writing unit tests for calculateNextTier covering:
   - Current value below the first tier
   - Current value equal to a tier
   - Current value above all tiers (maxed case)
   - Empty tiers array
3. Writing unit tests for calculatePercentage covering:
   - Current value below next tier
   - Current value equal to next tier
   - Current value above next tier
   - nextTier = null (maxed scenario)
4. Ensuring edge cases are properly handled to prevent regressions in achievement calculations.

I will make sure the tests are clean, maintainable, and consistent with the existing testing structure.

Kindly assign this issue to me if possible.

Thank you!

tmdeveloper007:
Closing as stale — no PR was opened for this issue. Please feel free to reopen if you'd like to work on this.

PR #2707, opened 2026-06-22
test: add unit tests for achievement-estimators helper functions #2695

desireddymohithreddy0925:
## Problem Statement
The helper functions `calculateNextTier` and `calculatePercentage` in `src/lib/achievement-estimators.ts` had no unit tests. Because they form the foundation of achievement tracking across the application, regressions here could incorrectly display user milestones and percentages.

## Proposed Solution
Export the pure, synchronous helper functions and add a comprehensive test suite to `test/achievement-estimators.test.ts` to ensure mathematically precise behavior under all conditions.

## Changes Made
- **Exported Helpers**: Added the `export` keyword to `calculateNextTier` and `calculatePercentage` in `src/lib/achievement-estimators.ts`.
- **Added `calculateNextTier` Tests**:
  - Covered when current value is below the first tier.
  - Covered when current value is exactly equal to a tier (moves to the next).
  - Covered maxed-out scenarios returning `null`.
  - Covered the edge case of an empty tiers array returning `null`.
- **Added `calculatePercentage` Tests**:
  - Covered the standard percentage calculation below the next tier.
  - Covered when the user exactly hits the next tier (returns `100`).
  - Covered when the user inexplicably exceeds the next tier (returns `100`).
  - Covered the maxed-out scenario where `nextTier` is `null` (returns `100`).

## Impact
This brings reliable test coverage to a critical part of the achievement logic, preventing any regressions in user progress tracking and ensuring accurate display across UI components.

### DIFF
--- a/src/lib/achievement-estimators.ts
+++ b/src/lib/achievement-estimators.ts
@@ -12,14 +12,14 @@ export interface AchievementEstimate {
 const TIERS_STANDARD = [1, 16, 128, 1024];
 const TIERS_STARSTRUCK = [16, 128, 512, 4096];
 
-function calculateNextTier(current: number, tiers: number[]): number | null {
+export function calculateNextTier(current: number, tiers: number[]): number | null {
   for (const tier of tiers) {
     if (current < tier) return tier;
   }
   return null; // Maxed out
 }
 
-function calculatePercentage(current: number, nextTier: number | null): number {
+export function calculatePercentage(current: number, nextTier: number | null): number {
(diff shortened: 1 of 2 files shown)


Priyanshu-byte-coder:
Closing as duplicate. Both this PR and #2726 add achievement-estimators tests but both also export previously private functions from production code to enable testing, which widens the public API surface unnecessarily. Please see the comment on #2726 for the requested approach. Thanks for contributing!

PR #2872, opened 2026-06-27
fix: cache GitHub API responses for compare endpoint to avoid rate limiting

### LINKED ISSUE #2841: [PERFORMANCE] GitHub API responses not cached server-side, causing repeated identical requests on every dashboard page load
## Problem

Each time the dashboard page loads, the server fires fresh API requests to GitHub for stats (commit counts, PR metrics, streak data) even when the underlying data has not changed. GitHub's REST API has a rate limit of 5,000 requests per hour per authenticated token. A user who refreshes the dashboard frequently, or a deployment with many active users sharing one token, will hit the rate limit and receive degraded responses (HTTP 403) until the window resets.

## Steps to Reproduce

1. Open the dashboard
2. Observe the network panel: GitHub API calls fire on every page load
3. Refresh the page 60-70 times in quick succession
4. Observe HTTP 403 responses with `X-RateLimit-Remaining: 0`

## Proposed Fix

Cache GitHub API responses server-side with a short TTL (5-15 minutes) using Redis or the existing Supabase instance:

```ts
const CACHE_TTL = 10 * 60; // 10 minutes in seconds

export async function getCachedGitHubStats(username: string) {
  const cached = await redis.get(`gh:stats:${username}`);
  if (cached) return JSON.parse(cached);
  const fresh = await fetchGitHubStats(username);
  await redis.setex(`gh:stats:${username}`, CACHE_TTL, JSON.stringify(fresh));
  return fresh;
}
```

Also display the last-refreshed timestamp in the UI so users understand they are viewing cached data.

**Complexity**: Level 2/3 | **Program**: GSSOC '26

Tanisha-sharma7302:
Fixes #2841

## Problem
The `/api/metrics/compare` endpoint had a TTL reserved for it in `metrics-cache.ts` 
(`compare: 30 * 60`) but never actually used it — instead it relied on a custom 
Supabase-based cache keyed per viewer + target user + calendar day. This meant 
every new (viewer, target, day) combination triggered 4-5 fresh GitHub API calls 
with no short-term caching, contributing to rate-limit (403) responses under 
repeated/rapid dashboard usage as described in the issue.

## Fix
Switched the compare endpoint to use the existing `withMetricsCache`-style 
Redis/memory caching system (`cacheGet`/`cacheSet` from `metrics-cache.ts`), 
using the already-reserved 30-minute TTL — consistent with every other metrics 
endpoint in the app. Also respects the existing `?refresh=true` / 
`?bypassCache=true` cache-bypass convention.

## Testing
- `pnpm run type-check` passes with no errors
- Existing test suite (`pnpm vitest run`) unaffected

### DIFF
--- a/src/app/api/metrics/compare/route.ts
+++ b/src/app/api/metrics/compare/route.ts
@@ -1,10 +1,16 @@
-import { getServerSession } from "next-auth";
+﻿import { getServerSession } from "next-auth";
 import { NextRequest } from "next/server";
 import { authOptions } from "@/lib/auth";
 import { toDateStr } from "@/lib/date-utils";
 import { calculateCurrentStreak } from "@/lib/streak";
 import { normalizeGitHubUsername } from "@/lib/validate-github-username";
-import { supabaseAdmin } from "@/lib/supabase";
+import {
+  cacheGet,
+  cacheSet,
+  isMetricsCacheBypassed,
+  metricsCacheKey,
+  METRICS_CACHE_TTL_SECONDS,
+} from "@/lib/metrics-cache";
 
 export const dynamic = "force-dynamic";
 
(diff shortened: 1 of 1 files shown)


Priyanshu-byte-coder:
Good caching improvement, but has BOM character at start of file and broken indentation (`const since90` loses leading spaces). Please fix formatting and rebase on main.

Priyanshu-byte-coder:
The caching refactor looks good in principle, but there is a blocking issue: the diff introduces a UTF-8 BOM (`\xEF\xBB\xBF`) at the start of `src/app/api/metrics/compare/route.ts` (the `+﻿import` line). Also, the indentation is inconsistent around `const since90 = new Date();` (missing leading spaces). Please remove the BOM and fix the indentation, then push an update.

Tanisha-sharma7302:
Fixed the BOM character and indentation issue in route.ts — pushed the update. Branch is already up to date with main. Ready for another review whenever you get a chance!

PR #2873, opened 2026-06-27
fix: detect revoked GitHub token immediately instead of waiting up to 24 hours

### LINKED ISSUE #2842: [BUG] GitHub OAuth access token expiry is not checked before API calls, causing silent 401 failures after token refresh
## Problem

The dashboard uses a GitHub OAuth access token stored in the user session. GitHub access tokens for OAuth apps do not expire by default, but fine-grained personal access tokens and GitHub Apps tokens do expire. Additionally, users can revoke authorization at any time in their GitHub settings. When an expired or revoked token is used to call the API, GitHub returns HTTP 401, but the dashboard does not distinguish this from a network error: it either shows stale data or a generic error message, leaving users confused about why their stats are missing.

## Steps to Reproduce

1. Authenticate with GitHub OAuth
2. Manually revoke the application's access in GitHub Settings > Applications
3. Reload the dashboard
4. Observe the generic error state rather than a "reconnect GitHub" prompt

## Proposed Fix

Inspect the HTTP status and `X-OAuth-Scopes` header on every GitHub API response:

```ts
if (response.status === 401) {
  await invalidateSession(userId);
  redirect("/auth/github"); // prompt re-authentication
}
```

Show a clear "Your GitHub connection has expired. Reconnect to refresh your stats." banner when this condition is detected.

Tanisha-sharma7302:
Fixes #2842

## Problem
The dashboard only re-validated the stored GitHub OAuth token once every 
24 hours (`TOKEN_VALIDATION_INTERVAL_MS` in `auth.ts`). If a user revoked 
the app's access in GitHub Settings, the app wouldn't notice until the next 
scheduled check — for up to 24 hours. In the meantime, every GitHub API call 
silently failed with 401, but the dashboard had no way to tell this apart 
from a generic network error, so it just showed stale data or a confusing 
error instead of prompting the user to reconnect.

The existing `TokenRevokedBanner` in `dashboard/layout.tsx` and the 
`window.fetch` 401-interceptor were already built to show a "reconnect" 
prompt — but they only fire when `session.error === "TokenRevoked"`, which 
was only ever set by the slow 24-hour periodic check.

## Fix
Added a fast path so a *live* 401 from GitHub immediately flags the token 
as revoked, instead of waiting for the periodic check:

- **`src/lib/token-revocation-flag.ts`** (new) — `markTokenRevokedNow()` / 
  `wasTokenRevokedNow()`, backed by the existing Redis cache with a 
  short TTL.
- **`src/lib/github-fetch.ts`** — when any GitHub API call returns a live 
  401, it now calls `markTokenRevokedNow()` for that user.
- **`src/lib/auth.ts`** — the NextAuth `jwt` callback now checks 
  `wasTokenRevokedNow()` on every request. If true, it sets 
  `session.error = "TokenRevoked"` immediately, instead of relying solely 
  on the 24-hour periodic check.

This means the existing `TokenRevokedBanner` ("Your GitHub session has 
expired. Please sign out and sign back in to refresh your data.") now 
shows up right away on the next page load after revocation, rather than 
up to a day later. The original 24-hour periodic check is left in place 
as a fallback for routes that don't actively call GitHub.

## Testing
- `pnpm run type-check` passes with no errors
- No existing tests were broken

### DIFF
--- a/src/lib/auth.ts
+++ b/src/lib/auth.ts
@@ -1,23 +1,17 @@
-import { type NextAuthOptions } from "next-auth";
+﻿import { type NextAuthOptions } from "next-auth";
 import GitHubProvider from "next-auth/providers/github";
 import { syncGitHubAchievementsForUser } from "./github-achievements";
 import { supabaseAdmin } from "./supabase";
+import { wasTokenRevokedNow } from "./token-revocation-flag";
 
 const SESSION_MAX_AGE = 30 * 24 * 60 * 60;
 const SESSION_UPDATE_AGE = 24 * 60 * 60;
 const isPlaywrightServer = process.env.PLAYWRIGHT_SERVER_MODE === "start";
 
 const GITHUB_API = "https://api.github.com";
-// Re-validate the stored GitHub token at most once every 24 hours per session.
-// Catches revocations within a reasonable window without adding per-request latency.
-// Without this check a revoked token silently continues for up to 30 days (JWT lifetime).
 const TOKEN_VALIDATION_INTERVAL_MS = 24 * 60 * 60 * 1000;
 
 export const authOptions: NextAuthOptions = {
-  // Playwright runs on plain HTTP (127.0.0.1) and relies on the default
-  // `next-auth.session-token` cookie name. If NextAuth infers HTTPS via
-  // forwarded headers, it may switch to secure cookie prefixes and the E2E
-  // session cookie won't be read. Force non-secure cookies in this mode.
   ...(isPlaywrightServer ? { useSecureCookies: false } : {}),
   providers: [
     GitHubProvider({
@@ -31,9 +25,6 @@ export const authOptions: NextAuthOptions = {
   pages: {
     signIn: "/auth/signin",
   },
-  // Use NextAuth's default cookie behavior (secure cookies on HTTPS deployments),
-  // which keeps Playwright E2E (http://127.0.0.1) aligned with the default
-  // `next-auth.session-token` cookie name.
   session: {
     strategy: "jwt",
     maxAge: SESSION_MAX_AGE,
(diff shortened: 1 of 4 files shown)


Priyanshu-byte-coder:
Good concept (fast-path revocation detection), but file has a BOM character at the start of auth.ts (`﻿`). Please remove the BOM and rebase on main.

Priyanshu-byte-coder:
Thanks for the token revocation fix — the logic looks useful. However, there is a blocking issue: the diff introduces a UTF-8 BOM character (`\xEF\xBB\xBF`) at the very start of `src/lib/auth.ts` (the `+﻿import` line). This will cause linting and possibly build failures. Additionally, the PR removes several explanatory comments that documented non-obvious behavior (e.g. why Playwright uses non-secure cookies, why the supabaseAdmin null guard exists). Please: 1) Remove the BOM from auth.ts, 2) Restore the removed doc-comments or replace with equivalent notes, then push an update.

Priyanshu-byte-coder:
The revocation-flag approach looks reasonable, but this diff strips out many load-bearing comments in src/lib/auth.ts (the Playwright cookie rationale, the supabaseAdmin null guard explanation, the 42703 fallback note) and introduces a BOM at the top of the file. Please restore the comments and remove the BOM, then we can re-review.

Priyanshu-byte-coder:
The feature here is genuinely good — a Redis-backed short-lived revocation flag so a live 401 surfaces `TokenRevoked` on the next session read instead of waiting for the 24h periodic check, and it's properly tested. Two things block it as-is:

1. **Byte-order marks (U+FEFF) got prepended to several files** — `src/lib/auth.ts`, `src/lib/github-fetch.ts`, `src/lib/token-revocation-flag.ts` and the test all start with a stray BOM (visible as `+﻿import`/`+﻿/**` in the diff). Please re-save those files as UTF-8 **without** BOM.
2. **The diff deletes a lot of existing explanatory comments** in `auth.ts` and `github-fetch.ts` (the Playwright cookie note, the supabaseAdmin-null guard rationale, the rate-limit classification notes, etc.). Those document non-obvious behavior — please keep them; this PR should *add* the fast-path, not strip the surrounding context.

Fix those two and it's a merge. Nice work on the flag design.

Priyanshu-byte-coder:
Closing as stale. The last commit on this branch was **2026-06-27**, over a month ago, and there's been no activity since. 

This isn't a rejection of the idea — DevTrack's PR queue grew past 110 open PRs, so I'm clearing branches that have gone quiet so the review tab reflects what's actually in progress.

If you still want to land this: rebase on current `main`, address any review feedback above, and reopen this PR (or open a fresh one) — I'll review it. Thanks for contributing to DevTrack.

Issue #2804, opened 2026-06-25
test : add unit tests for csv.ts


tmdeveloper007:
Summary of What Needs to be Done: Add unit tests for the pure CSV serialization utilities exported from src/lib/csv.ts: the csvCell() and toCsv() functions.

Changes that Need to be Made: Create test/csv.test.ts using vitest. Test csvCell() with: null and undefined (expect empty string), plain text without special chars (expect unquoted string), text containing a comma (expect quoted), text containing a double-quote (expect escaped with doubled quote, wrapped in quotes), text containing newline or carriage-return (expect quoted). Test toCsv() with: empty array (expect empty string), single row with all headers, multiple rows with consistent and missing keys, numeric and boolean values (auto-coerced to string without quoting).

Impact that it would Provide: Provides test coverage for the CSV export utility used in data export features. Ensures RFC 4180-compliant quoting and escaping behavior is maintained. Guards against regressions as export functionality expands.

Tuba1809:
Hii, I would like to work on the issue.
Please assign it to me!!

yachikadev:
hii i would like to work on this issue contributing under gssoc'26

tmdeveloper007:
Closing this issue as it has been open for more than 2 days without activity.

PR #2877, opened 2026-06-27
test: fix unit tests for csv.ts (csvCell and toCsv)

### LINKED ISSUE #2941: test : add unit tests for csv.ts
### Bug Description

> **Note:** This issue is being reopened. The original issue #2804  was closed by the owner, but I found the proposed changes helpful and have already submitted a PR for the same. Hence, I am reopening this to track the work.

Summary of What Needs to be Done: Add unit tests for the pure CSV serialization utilities exported from src/lib/csv.ts: the csvCell() and toCsv() functions.

Changes that Need to be Made: Create test/csv.test.ts using vitest. Test csvCell() with: null and undefined (expect empty string), plain text without special chars (expect unquoted string), text containing a comma (expect quoted), text containing a double-quote (expect escaped with doubled quote, wrapped in quotes), text containing newline or carriage-return (expect quoted). Test toCsv() with: empty array (expect empty string), single row with all headers, multiple rows with consistent and missing keys, numeric and boolean values (auto-coerced to string without quoting).

Impact that it would Provide: Provides test coverage for the CSV export utility used in data export features. Ensures RFC 4180-compliant quoting and escaping behavior is maintained. Guards against regressions as export functionality expands.

yachikadev:
## Summary

Fixes the existing unit tests in `test/csv.test.ts` for CSV serialisation 
utilities exported from `src/lib/csv.ts`.

Closes #2941 

## Changes Made

- Fixed incorrect expectation in `toCsv` single-row test case
  - Previous: expected header row only (`"name,age"`)
  - Fixed: expects header + data row (`"name,age\nAlice,30"`) which matches 
    actual implementation behaviour

## Why

The existing test assumed `toCsv` with a single row returns only headers. 
But the implementation always appends data rows — even for a single row. 
The wrong expectation would have caused CI to fail.

## Test Coverage (already present, now correct)

All requirements from issue #2804 are covered:
- `csvCell`: null/undefined, plain text, comma, double-quote, newline, carriage return
- `toCsv`: empty array, single row, multiple rows, missing keys, numbers/booleans

### DIFF
--- a/test/csv.test.ts
+++ b/test/csv.test.ts
@@ -54,8 +54,8 @@ describe("toCsv", () => {
     expect(toCsv([])).toBe("");
   });
 
-  it("returns header row only for single row", () => {
-    expect(toCsv([{ name: "Alice", age: 30 }])).toBe("name,age");
+  it("returns header and data row for single row", () => {
+  expect(toCsv([{ name: "Alice", age: 30 }])).toBe("name,age\nAlice,30");
   });
 
   it("serialises multiple rows with correct values", () => {


yachikadev:
The Playwright smoke test failure is not related to this PR.
It is caused by a pnpm-lock.yaml lockfile mismatch with the current overrides config in package.json — a pre-existing repo issue. All checks relevant to this PR pass successfully.

yachikadev:
@Priyanshu-byte-coder Hi! I noticed the issue was closed — could you let me know if this PR is still relevant or should I close it? Happy to make any changes if needed.

PR #2875, opened 2026-06-27
fix(security): remove GitHub access token from client-exposed session

BCA-krishna:
Closes #2845

## Problem
The NextAuth `session()` callback copied the GitHub access token onto
the `session` object. Since `session` is exposed to client-side code via
`useSession()`/`getSession()`, the raw GitHub token was readable by any
JavaScript running on the page — including a successful XSS payload —
giving full access to the user's GitHub account with the granted scopes.

## Fix
- Removed `session.accessToken = token.accessToken` from the `session()`
  callback in `src/lib/auth.ts`. The token now stays only in the
  encrypted, HttpOnly JWT cookie that NextAuth manages — it never reaches
  the browser.
- Added a `getAccessToken()` helper in `src/lib/get-session-token.ts`
  that reads the token **server-side only**, directly from the JWT via
  `next-auth/jwt`'s `getToken()`.
- Migrated all API routes that previously read `session.accessToken` to
  call `getAccessToken()` instead (29 route files).
- Updated/added tests to mock `getAccessToken()` and to assert that
  `accessToken` is never present on the session object returned to the
  client.

## Testing
- `npm run type-check` — 0 errors
- `npm run lint` — 0 errors (5 pre-existing unrelated `` warnings)
- `npm test` — same failure count as `main` (all pre-existing/unrelated
  to this change — verified via `git stash` comparison); no new failures
  introduced by this fix

PR #3132, opened 2026-07-04
fix(security): stop exposing GitHub access token on client-visible session (#2845)

### LINKED ISSUE #2845: [SECURITY] GitHub personal access tokens stored in plaintext browser localStorage are accessible to any JavaScript running on the page
## Problem

If the dashboard stores the GitHub OAuth access token or personal access token in `localStorage`, it is accessible to any JavaScript running on the same origin, including third-party scripts, browser extensions, and any XSS payload. This allows a successful XSS attack to silently exfiltrate the token and gain full access to the user's GitHub account with the granted scopes.

## Steps to Reproduce

1. Authenticate with GitHub
2. Open the browser console and run: `localStorage.getItem("gh_token")` (or check    all localStorage keys for any token-like value)
3. If a token is found, confirm it is a valid GitHub token by calling    `https://api.github.com/user` with it as a Bearer token

## Proposed Fix

- Store tokens exclusively in HTTP-only, Secure, SameSite=Strict cookies set by   the server. These are invisible to JavaScript and safe from XSS exfiltration
- If client-side token storage is required, use `sessionStorage` with a short expiry   and rotate tokens frequently
- Audit all `localStorage.setItem()` calls to confirm no sensitive values are stored there

**Complexity**: Level 3 | **Program**: GSSOC '26

BCA-krishna:
Closes #2845

## What changed and why

The NextAuth `session()` callback in `src/lib/auth.ts` was copying the
raw GitHub OAuth access token onto the `session` object. Since `session`
is returned to the browser via `useSession()`/`getSession()`, any
JavaScript on the page — including an XSS payload — could read the token
and gain full GitHub API access with the granted scopes.

## Files changed (3 files only)

- **`src/lib/auth.ts`** — removed `session.accessToken = token.accessToken`
  from the `session()` callback. Token stays in the encrypted HttpOnly
  JWT cookie only.
- **`src/lib/get-session-token.ts`** — added `getAccessToken()`, a
  `server-only` helper that reads the token directly from the JWT via
  `next-auth/jwt`'s `getToken()`, for use in API route handlers.
- **`test/auth.test.ts`** — updated test to assert that `accessToken` is
  NOT present on the session object (previously asserted the opposite).

## Follow-up

API routes that currently read `session.accessToken` will be migrated to
`getAccessToken()` in a follow-up PR. Those routes are all server-side
(`src/app/api/`) and never reach the browser, so they continue to work
correctly via the session callback until migrated.

## Testing
- `npm run type-check` — 0 errors
- `npm run lint` — 0 errors
- `npm test` — auth.ts tests all pass (18/18)

### DIFF
--- a/src/lib/auth.ts
+++ b/src/lib/auth.ts
@@ -175,8 +175,6 @@ export const authOptions: NextAuthOptions = {
       return token;
     },
     async session({ session, token }) {
-      if (typeof token.accessToken === "string")
-        session.accessToken = token.accessToken;
       if (typeof token.githubId === "string")
         session.githubId = token.githubId;
       if (typeof token.githubLogin === "string")
--- a/src/lib/get-session-token.ts
+++ b/src/lib/get-session-token.ts
@@ -1,5 +1,7 @@
 import "server-only";
 import { getServerSession } from "next-auth";
+import { getToken } from "next-auth/jwt";
+import { headers, cookies } from "next/headers";
 import { authOptions } from "@/lib/auth";
 import { resolveAppUser } from "@/lib/resolve-user";
 import type { Session } from "next-auth";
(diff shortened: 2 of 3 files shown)


BCA-krishna:
Hi @Priyanshu-byte-coder , please review this PR.
thanks.

Priyanshu-byte-coder:
Same author, minimal version of #2875 — but removing `session.accessToken` on its own can't work: 39 API routes still read `session.accessToken`, so this branch red-lights the Playwright smoke tests (all those routes lose their token). The full migration in #2875 is the way to land this. Closing here in favor of getting #2875 fixed up (it has a Next 16 async `headers()`/`cookies()` bug to sort out first — details on that PR). Let's consolidate the effort there.

PR #3133, opened 2026-07-04
fix(repos): paginate repo-explorer to fetch all repos beyond 100 limit (#2843)

### LINKED ISSUE #2843: [BUG] Repository list fetches all repos in a single API call with no pagination, causing failures for users with more than 100 repositories
## Problem

The GitHub REST API returns a maximum of 100 repositories per page. The current implementation calls `/user/repos?per_page=100` once and uses only that first page. Users who have more than 100 repositories (common for active developers) see an incomplete repo list on the dashboard. Pull request metrics and streak calculations derived from that list are therefore also incomplete.

## Steps to Reproduce

1. Authenticate as a user with more than 100 GitHub repositories
2. Open the dashboard and view the repository list
3. Count the displayed repos and compare with the GitHub profile count
4. Confirm repos beyond 100 are missing

## Proposed Fix

Paginate through all pages until a response returns fewer than `per_page` items:

```ts
async function fetchAllRepos(token: string) {
  const repos = [];
  let page = 1;
  while (true) {
    const batch = await fetchGitHubRepos(token, page, 100);
    repos.push(...batch);
    if (batch.length < 100) break;
    page++;
  }
  return repos;
}
```

Cache the result to avoid N paginated requests on every page load.

BCA-krishna:
Closes #2843

## Problem
The repo-explorer endpoint called `/user/repos?per_page=100` once,
returning at most 100 repositories. Users with more than 100 repos
saw an incomplete list on the dashboard, and metrics derived from
that list (PR counts, streak) were also incomplete.

## Fix
- Replaced the single `fetch()` call in `repo-explorer/route.ts` with
  the existing `fetchUserRepos()` helper from `src/lib/github.ts`, which
  already paginates up to 10 pages (1000 repos maximum).
- Added missing `created_at` and `language` fields to the `GitHubRepo`
  interface so the type accurately reflects the GitHub API response.
- Updated test mock objects to include the new required fields.

## Testing
- `npm run type-check` — 0 errors
- `npm run lint` — 0 errors

### DIFF
--- a/src/app/api/metrics/repo-explorer/route.ts
+++ b/src/app/api/metrics/repo-explorer/route.ts
@@ -1,4 +1,5 @@
 import { getSessionWithToken } from "@/lib/get-session-token";
+import { fetchUserRepos } from "@/lib/github";
 import { NextRequest } from "next/server";
 import { isMetricsCacheBypassed, metricsCacheKey, withMetricsCache } from "@/lib/metrics-cache";
 import { ExplorerRepoCardData } from "@/lib/repo-analytics-types";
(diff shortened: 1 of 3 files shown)


BCA-krishna:
Hi @Priyanshu-byte-coder , please make review in my PR.
Thanks.

PR #3157, opened 2026-07-06
feat: add floating scroll-to-top button for Agents list (#3149)

### LINKED ISSUE #3149: [FEAT] Add a Floating "Scroll to Top" Button for the Agents List
### Problem Statement

As a user, I want to quickly return to the top of the Agents list so that I can switch providers, use the search bar, or navigate without manually scrolling through a long list.

As the number of available agents continues to grow, returning to the top of the page becomes increasingly time-consuming. A dedicated "Scroll to Top" button would improve navigation and overall usability.

### Proposed Solution

Add a floating "Scroll to Top" button that appears in the bottom-right corner after the user scrolls beyond a predefined distance (e.g., 300–500px).

### Behavior
- The button remains hidden when the page loads.
- It becomes visible once the user scrolls past the threshold.
- Clicking the button smoothly scrolls the page back to the top.
- The button should be accessible with an appropriate `aria-label`.
- It should be responsive and work well on both desktop and mobile devices.
- The button should not obstruct existing UI elements.

### Feature Area

Dashboard

### Alternatives Considered

- Relying on the browser's scrollbar or keyboard shortcuts.
- Keeping a permanently visible button, which adds unnecessary UI clutter.
- Using only a keyboard shortcut, which is not discoverable for all users.

A floating button that appears only when needed offers the best balance between usability and a clean interface.

### Acceptance Criteria

- [ ] A floating "Scroll to Top" button appears after the user scrolls beyond the configured threshold.
- [ ] Clicking the button smoothly scrolls the page to the top.
- [ ] The button is hidden when the user is near the top of the page.
- [ ] The button is responsive and works correctly on desktop and mobile.
- [ ] The button includes appropriate accessibility attributes (e.g., `aria-label`).
- [ ] The feature does not interfere with existing UI components.

### Additional Context

Many modern applications and documentation websites include a floating "Scroll to Top" button to improve navigation on long pages. This is a lightweight UX enhancement that requires minimal implementation while providing a noticeable improvement in usability.

saksham242007:
Adds a floating "Scroll to Top" button on the dashboard that appears after scrolling past 400px and smoothly scrolls back to top when clicked.

Closes #3149

### DIFF
--- a/src/app/dashboard/page.tsx
+++ b/src/app/dashboard/page.tsx
@@ -15,6 +15,7 @@ import { redirect } from "next/navigation";
 import DashboardSSEProvider from "@/components/DashboardSSEProvider";
 import { DashboardWidgetA11yProvider } from "@/components/dashboard/DashboardWidgetA11yContext";
 import RoastHypeWidget from "./RoastHypeWidget";
+import ScrollToTopButton from "@/components/ScrollToTopButton";
 
 
 export default async function DashboardPage() {
@@ -90,7 +91,7 @@ export default async function DashboardPage() {
 
             {/* Today Focus */}
             <section>
-              <TodayFocusHero userName={session.user?.name ?? null} />
+              <TodayFocusHero userName={session?.user?.name ?? null} />
             </section>
 
             {/* Featured Section */}
@@ -146,6 +147,8 @@ export default async function DashboardPage() {
               <MilestonePlanner />
             </section>
           </div>
+
+          <ScrollToTopButton />
         </main>
       </DashboardWidgetA11yProvider>
     </DashboardSSEProvider>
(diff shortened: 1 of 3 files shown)


Priyanshu-byte-coder:
DevTrack already has a floating back-to-top control — `BackToTopButton`, rendered app-wide in `src/app/providers.tsx`, so it's already on the dashboard (and every page), and it's a bit richer than this one (it shows a circular scroll-progress ring). Adding `ScrollToTopButton` to the dashboard would put two back-to-top buttons on the same page. So I can't take this as a new component. Also note the diff carries an unrelated blank line in `middleware.ts` and a `session.user`→`session?.user` tweak that don't belong here. If you'd like to improve scroll-to-top, a PR that enhances the existing `BackToTopButton` (e.g. on the Agents list specifically, per #3149) would be welcome. Closing this one.

PR #3173, opened 2026-07-08
fix(security): prevent TOCTOU DNS rebinding SSRF in webhooks

### LINKED ISSUE #3169: [BUG] TOCTOU DNS Rebinding Vulnerability in Webhook Dispatch
### Bug Description

The application allows outbound webhooks. To prevent Server-Side Request Forgery (SSRF), it checks `isSafeUrl()`, which performs a DNS lookup to ensure the IP isn't internal (e.g., `127.0.0.1`, AWS Metadata `169.254.169.254`). However, after passing the check, `fetch(webhook.url)` is called. This forces a *second* DNS resolution. 

This introduces a classic Time-of-Check to Time-of-Use (TOCTOU) vulnerability. An attacker can set up a DNS server with an extremely low TTL. The first request (the check) returns a safe, public IP. The second request (the `fetch`) returns an internal IP (like `169.254.169.254`). The attacker can then exfiltrate the serverless container's IAM roles, internal network topology, or hit internal databases.

### Steps to Reproduce

1. Register an outbound webhook with a URL pointing to an attacker-controlled DNS server (e.g., `http://rebind.attacker.com`).
2. Configure the DNS server to return `8.8.8.8` on the first resolution, and `169.254.169.254` on the second resolution.
3. Trigger the webhook dispatch.
4. Observe that the SSRF check passes, but the subsequent fetch hits the internal AWS Metadata IP or local services.

### Affected Area
API Routes

### Screenshots
_No response_

### Browser & OS
_No response_

### Environment
Both

### Additional Context

**Suggested Fix:**
1. Modify `isSafeUrl` to return the resolved, safe IP address instead of a boolean.
2. Pass this exact IP address directly to the `fetch` call.
3. Override the `Host` header in the `fetch` options to match the original URL's hostname so the target server processes it correctly, ensuring the IP used is explicitly pinned.

nyxsky404:
This PR fixes a critical TOCTOU (Time-of-Check to Time-of-Use) vulnerability in the webhook dispatch system. Previously, `isSafeUrl` performed a DNS resolution to verify the IP, but `fetch` subsequently performed a second DNS resolution. This allowed attackers to use DNS rebinding (with a low TTL) to bypass SSRF protections and hit internal networks (e.g. 169.254.169.254 or localhost). The fix updates `isSafeUrl` to return the explicitly validated IP address, which is then passed directly to `fetch` while pinning the `Host` header to the original URL's hostname. Fixes #3169.

### DIFF
--- a/src/app/api/metrics/repo-analytics/route.ts
+++ b/src/app/api/metrics/repo-analytics/route.ts
@@ -37,7 +37,7 @@ export async function GET(req: NextRequest) {
   const repoUrl = `${GITHUB_API}/repos/${safeRepoPath}`;
   let urlSafe = false;
   try {
-    urlSafe = await isSafeUrl(repoUrl);
+    urlSafe = (await isSafeUrl(repoUrl)).safe;
   } catch {
     urlSafe = false;
   }
--- a/src/app/api/webhooks/custom/[id]/deliveries/[deliveryId]/retry/route.ts
+++ b/src/app/api/webhooks/custom/[id]/deliveries/[deliveryId]/retry/route.ts
@@ -67,7 +67,7 @@ export async function POST(
     .eq("id", id)
     .single();
 
-  if (!webhookUrl || !(await isSafeUrl(webhookUrl.url))) {
+  if (!webhookUrl || !(await isSafeUrl(webhookUrl.url)).safe) {
     return Response.json(
       { error: "Webhook URL is not allowed. Private, loopback, and internal addresses are blocked." },
       { status: 400 }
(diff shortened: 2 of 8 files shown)


Priyanshu-byte-coder:
Solid, thorough change and CI is green — but there's a runtime defect the tests don't catch. In `dispatchWebhook` you set `fetchUrl.hostname = ip` and then call `fetch()`. For **HTTPS** webhooks that breaks TLS: undici connects to the IP, uses the IP as the SNI servername, and validates the server certificate against the IP rather than the original hostname — so real HTTPS deliveries fail with `ERR_TLS_CERT_ALTNAME_INVALID`. The unit test mocks `fetch`, so it passes anyway. Correctly pinning the resolved IP while keeping TLS valid needs a raw `https.request` with `servername` set to the original hostname (SNI) plus a `Host` header — which is exactly what #2669 does. Closing in favor of consolidating the TOCTOU fix there. Thanks for the careful work on the call-site refactor.

PR #3190, opened 2026-07-12
perf: paginate discord-sync + push filters into SQL (#3163)

### LINKED ISSUE #3163: [Performance] /api/notifications/discord-sync fetches every discord user in a single query — cron truncates
### Description

similar shape of bug to the wakatime cron, the discord notification cron loads every user with a discord webhook in a single unbounded query.

### The Problem

src/app/api/notifications/discord-sync/route.ts:21-24:

```
  .from("users")
  .select("id, github_login, discord_webhook_url, timezone, ...")
  .not("discord_webhook_url", "is", null);
```

no `.range()`, no `.limit()`, no ordering. line 36 then serially loops with `for (const user of users)` and for each one calls `fetchPublicStreak` and possibly `fetchPublicContributions` (github API roundtrips) plus a supabase update. so the query size AND the loop time both scale with total discord users.

failure scenario: this cron runs hourly. once discord users cross a few hundred, the outer loop cannot finish inside the vercel function timeout and the whole batch is truncated. users at the end of the query never get notifications. worse, the initial supabase read pulls all rows including muted / already-notified users that get skipped later at lines 60-77 anyway, so most of the payload is wasted.

### Proposed Fix

- page the users query in batches of 50 same as src/app/api/cron/sync/route.ts:38-128
- push the "which users need a notification this hour" filters into the SQL where clause: filter on `last_discord_notification_at < now() - interval '20 hours'` and `discord_muted_until is null or discord_muted_until < now()` so we do not fetch users we will just skip
- optionally split streak-at-risk vs weekly-summary into separate scheduled runs so the sunday load does not stack

I would like to work on this under GSSoC 26. please assign!

for labels i think gssoc + level2 + type:performance fits here.

vedant7007:
### Summary

\`/api/notifications/discord-sync\` did one unbounded \`.select()\` and pulled every user with a discord webhook (webhook URL, timezone, mute state, last notification timestamp) into serverless memory in a single response — then serially walked each through 2-3 GitHub API roundtrips + a Supabase update. Both scaled with total user count, so once discord users cross a few hundred the outer loop can't finish within the Vercel function timeout and users at the tail get silently skipped every hour.

### What changed

- Adopted the pagination pattern already used by \`/api/cron/sync\` (\`PAGE_SIZE=50\`, deterministic \`.order("id")\`, \`.range(page*50, page*50+49)\`, break out when a short page returns).
- Pushed the "hasn't been notified in the last 20 hours" and "not currently muted" filters into the SQL \`WHERE\` clause via \`.or()\` so we don't fetch rows we'll immediately skip. Cuts the wire payload per page down to actual candidates.
- Cached \`twentyHoursAgo\` and \`nowIso\` outside the loop.

### What's unchanged

The per-user \`localHour\` / weekday / streak-at-risk / weekly-summary / milestone / notification-sending logic is identical to the old code path. No behavior change for notification decisions — only the fetch/scan strategy differs.

### Test plan

- [x] \`tsc --noEmit\` clean
- [x] verified the \`.or()\` clauses match the previous in-JS filter logic (NULL or condition satisfied)
- [ ] once approved: manual smoke with a couple of test users to confirm pagination unwinds correctly on both an empty tail and a full page

Fixes #3163

### DIFF
--- a/src/app/api/notifications/discord-sync/route.ts
+++ b/src/app/api/notifications/discord-sync/route.ts
@@ -18,112 +18,136 @@ export async function GET(req: Request) {
   const authError = validateCronRequest(req);
   if (authError) return authError;
 
-  const { data: users, error } = await supabaseAdmin
-    .from("users")
-    .select("id, github_login, discord_webhook_url, timezone, last_discord_notification_at, discord_muted_until")
-    .not("discord_webhook_url", "is", null);
-
-  if (error || !users) {
-    return NextResponse.json({ error: "Database error" }, { status: 500 });
-  }
-
   const token = process.env.GITHUB_TOKEN;
   const now = new Date();
 
   let processed = 0;
   let notificationsSent = 0;
 
-  for (const user of users) {
-    if (!user.discord_webhook_url) continue;
-    if (!isValidDiscordWebhookUrl(user.discord_webhook_url)) continue;
+  // Paginate the users query. The old implementation did one unbounded
+  // `.select()` and pulled every discord user (webhook URL, mute status,
+  // etc.) into serverless memory in a single response, then serially
+  // walked each one through 2-3 GitHub API roundtrips. Both scaled with
+  // total user count, so once discord users cross a few hundred the outer
+  // loop couldn't finish within the Vercel function timeout and users at
+  // the tail were silently skipped. Mirror the pattern used by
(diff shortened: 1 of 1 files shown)


vedant7007:
starred the repo — should turn the star-required gate green on the next re-run 🌟

PR #3203, opened 2026-07-13
perf: paginate wakatime/sync users query (#3161)

### LINKED ISSUE #3161: [Performance] /api/wakatime/sync loads every user with a wakatime key in one unbounded query
### Description

while comparing the wakatime cron route with the newer /api/cron/sync one i noticed the wakatime one still does an unbounded fetch of the entire users table.

### The Problem

src/app/api/wakatime/sync/route.ts:13-17:

```
  .from("users")
  .select("id, wakatime_api_key_encrypted, wakatime_api_key_iv")
  .not("wakatime_api_key_encrypted", "is", null)
  .not("wakatime_api_key_iv", "is", null);
```

no `.range()`, no `.limit()`. every wakatime user in the DB (plus their encrypted keys) is pulled into serverless memory in one shot. line 29 then loops through `users.length` in chunks of 5.

failure scenario: on a real deployment with a few thousand wakatime users this either hits the vercel 4.5mb response limit from supabase, blows the 10s/60s function timeout, or gets killed by the runtime memory cap and the entire nightly sync silently fails. same class of bug that was already fixed in src/app/api/cron/sync/route.ts:38-128 which pages with `PAGE_SIZE = 50` and a `while (hasMore)` loop.

### Proposed Fix

- adopt the same pagination pattern as cron/sync: page through 50 users at a time using `.range(page * PAGE_SIZE, (page + 1) * PAGE_SIZE - 1)`
- break out of the while loop when the returned page is smaller than `PAGE_SIZE`
- keep the existing `Promise.allSettled` chunk-of-5 inside each page
- also add `.order("id")` so pagination is deterministic

I would like to work on this under GSSoC 26. please assign!

for labels i think gssoc + level2 + type:performance fits here.

vedant7007:
### Summary

\`/api/wakatime/sync\` did one unbounded \`.select()\` and pulled every user with a wakatime API key (plus their encrypted keys) into serverless memory in a single response. At a few thousand users that would either:
- hit Supabase's ~4.5 MB response cap,
- blow the Vercel function timeout, or
- get killed by the runtime memory limit,

and the nightly sync would silently fail.

### What changed

Adopted the same pagination pattern already in use by \`/api/cron/sync\`:

- \`PAGE_SIZE = 50\`
- deterministic \`.order("id")\` so pagination doesn't drift
- \`.range(page * PAGE_SIZE, (page + 1) * PAGE_SIZE - 1)\`
- \`while (hasMore)\` loop that breaks out when a short page is returned

Kept the inner \`Promise.allSettled\` \`CHUNK_SIZE = 5\` for parallelism across the Wakatime API within each page — that part was already sensible, just needed to be nested inside the pagination.

### What's unchanged

Decrypt-then-fetch-then-upsert logic per user is identical. No behavior change for the actual sync — only the fetch/scan strategy differs.

### Test plan

- [x] \`tsc --noEmit\` clean
- [x] verified fix mirrors the sibling \`/api/cron/sync\` pagination line-for-line (same PAGE_SIZE, same \`.order("id")\` + \`.range()\` shape, same \`if (users.length < PAGE_SIZE) hasMore = false\` tail check)
- [ ] once approved: manual smoke — confirm the cron completes across a few paginated batches without truncating

Fixes #3161

### DIFF
--- a/src/app/api/wakatime/sync/route.ts
+++ b/src/app/api/wakatime/sync/route.ts
@@ -9,80 +9,106 @@ export async function GET(req: Request) {
   const authError = validateCronRequest(req);
   if (authError) return authError;
 
-  // Fetch users with wakatime keys
-  const { data: users, error } = await supabaseAdmin
-    .from("users")
-    .select("id, wakatime_api_key_encrypted, wakatime_api_key_iv")
-    .not("wakatime_api_key_encrypted", "is", null)
-    .not("wakatime_api_key_iv", "is", null);
-
-  if (error) {
-    console.error("Failed to fetch users for wakatime sync:", error);
-    return NextResponse.json({ error: "Database error" }, { status: 500 });
-  }
-
+  // Paginate the users query. The old implementation did one unbounded
+  // `.select()` and pulled every wakatime user (plus their encrypted keys)
+  // into serverless memory in a single response. At a few thousand users
+  // that either hits Supabase's ~4.5 MB response cap, blows the function
+  // timeout, or gets killed by the memory limit and the nightly sync
+  // silently fails. Mirror the pattern used by /api/cron/sync (which was
+  // already paginated) — 50-user pages, deterministic ordering by id,
+  // break out when the page is short. See issue #3161.
+  const PAGE_SIZE = 50;
+  const CHUNK_SIZE = 5;
   let successCount = 0;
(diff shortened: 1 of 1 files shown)


vedant7007:
starred the repo — should turn the star-required gate green on the next re-run 🌟

PR #3237, opened 2026-07-25
Fix inconsistent timezone handling in contributions API and StreakCalendar

saurabhhhcodes:
### Bug 1: Mixed timezone handling in contributions API route

`fetchContributionsForAccount` used `toLocalDateStr()` (local timezone) to format the `since` date for the GitHub API query, while every other route uses `toISOString().slice(0, 10)` (UTC). For users in negative UTC offsets this caused a one-day shift in the query window, making streak and contribution data disagree on which dates to include. Fixed by removing `toLocalDateStr` and using `since.toISOString().slice(0, 10)` consistently with the rest of the codebase.

### Bug 2: StreakCalendar used local dates to look up UTC-keyed contribution data

The `StreakCalendar` component used `toLocalDateStr()` (local timezone) to construct the lookup key into the contribution data, but the data keys come from GitHub's API which always returns UTC dates (`item.commit.author.date.slice(0, 10)`). This caused commits to appear on wrong calendar cells for users whose local date differs from UTC. Fixed by renaming to `toUtcDateStr` and using `toISOString().slice(0, 10)` for both lookup and isToday comparison.

### DIFF
--- a/src/app/api/metrics/contributions/route.ts
+++ b/src/app/api/metrics/contributions/route.ts
@@ -17,10 +17,6 @@ interface ContributionResponse {
   data: Record<string, number>;
 }
 
-function toLocalDateStr(d: Date): string {
-  return `${d.getFullYear()}-${String(d.getMonth() + 1).padStart(2, "0")}-${String(d.getDate()).padStart(2, "0")}`;
-}
-
 function mergeContributionDays(
   a: Record<string, number>,
   b: Record<string, number>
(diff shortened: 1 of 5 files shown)


Priyanshu-byte-coder:
This one is genuine work and I want it — unlike the bulk PRs, this fixes real bugs:

1. **Merged-PR count was wrong.** `data.items.filter(pr => pr.state === "closed")` counted *every* closed PR as merged, so closed-unmerged PRs inflated the number. Your `pr.pull_request?.merged_at != null` is the correct check (and correctly uses `!=` so it catches `undefined` too).
2. **Timezone consistency** — swapping the local-date helper for `toISOString().slice(0,10)` matches GitHub's UTC commit dates.
3. **Pagination** via the `link` header with `Promise.allSettled`, capped at 5 pages.

Two things before it can merge:

- **It conflicts with `main`.** The branch is based on a much older commit (its `ci.yml` still uses npm and `actions/checkout@v4`), so please rebase onto current `main` and resolve.
- **Concurrency:** the pagination fires up to 4 parallel Search API calls per user. GitHub's guidance is to make requests for a single user *serially* — Search has a low secondary rate limit. Please cap it to 2 concurrent, or fetch sequentially.

Rebase + that tweak and I'll merge it. This is the standard of work I want from you — please put your effort here rather than into bulk submissions.

PR #3273, opened 2026-07-29
fix: add .catch() to promise chains

saurabhhhcodes:


Priyanshu-byte-coder:
Closing — the change is incorrect. It moves the radix argument into the wrong function:

```diff
- parseInt(req.nextUrl.searchParams.get("days") ?? "90", 10)
+ parseInt(req.nextUrl.searchParams.get("days", 10) ?? "90", 10)
```

`URLSearchParams.get()` takes exactly one argument, so this breaks the build (`TS2554: Expected 1 arguments, but got 2`). And it wasn't needed — `parseInt` already had its radix `10` as the second argument, which is the correct place for it.

Note the title doesn't match the diff either. Please run `pnpm type-check` locally before pushing.

PR #3279, opened 2026-07-29
fix: add radix parameter to parseInt calls

saurabhhhcodes:


Priyanshu-byte-coder:
Closing — this is a re-open of #3274, which I closed on 2026-07-29 explaining the problem. The change moves the radix into the wrong function:

```diff
- parseInt(req.nextUrl.searchParams.get("days") ?? "90", 10)
+ parseInt(req.nextUrl.searchParams.get("days", 10) ?? "90", 10)
```

`URLSearchParams.get()` takes one argument, so this breaks the build, and `parseInt` already had its radix in the correct position. This version additionally deletes 171 lines. Please don't re-open changes that were closed with an explanation.

PR #3299, opened 2026-07-31
fix: code quality and safety improvements

saurabhhhcodes:


Priyanshu-byte-coder:
Closing — this change is incorrect and would introduce a bug.

```diff
- return Array.from(reposSet).sort();
+ return Array.from(reposSet).sort((a, b) => a - b);
```

`reposSet`, `langsSet` and `events` all hold **strings**. `(a, b) => a - b` is a *numeric* comparator — subtracting two strings yields `NaN`, so the comparator always returns `NaN` and the sort order becomes unspecified. The default `.sort()` (lexicographic) is the correct behavior for strings.

This matters beyond ordering: `eventsKey = [...events].sort().join(",")` is used as a cache key, so a non-deterministic sort would break cache stability.

Also note #3299 and #3300 are identical to each other — please don't open the same change twice.

PR #3314, opened 2026-07-31
Refactor dashboard widget tests for better clarity

### LINKED ISSUE #3313: [BUG] Fix route shadowing wildcard breaking goals sync and remove redundant click
### Bug Description

There are two distinct issues in this test file:

    Route Shadowing: The wildcard route "**/api/goals**" is registered after "**/api/goals/sync**". Because Playwright evaluates routes in reverse order of registration (last registered takes precedence), the broader wildcard pattern incorrectly intercepts POST requests meant for /api/goals/sync. It returns a generic 201 Created with { ok: true } instead of the expected payload containing last_synced_at, causing the dashboard's sync logic to fail or crash.

    Redundant/Flaky Action: In the contribution graph test, there is a duplicated click action: a blind .first().click() on a button, followed immediately by a correctly scoped .locator(...).click() on the same button name. This is a leftover strict-mode workaround that can cause race conditions or unintended UI state changes (e.g., clicking the wrong button in a different widget before clicking the correct one).

### Steps to Reproduce

Run the Playwright test suite for the dashboard.

The beforeEach hook mounts the API mocks. **/api/goals/sync** is registered, but then shadowed by the overly greedy **/api/goals** wildcard.

The frontend application automatically dispatches a POST request to /api/goals/sync.

The wildcard catches it, stripping the last_synced_at field from the response and breaking the frontend's expected data contract.

In the contribution graph test, the runner blindly clicks the first instance of "Show 90-day range" (which could belong to an entirely different graph if responsive layouts render multiple), risking test flakiness, before clicking the correctly scoped one.

### Affected Area

Other

### Screenshots

_No response_

### Browser & OS

_No response_

### Environment

None

### Additional Context

_No response_

rudrakshtank:
## Summary

Closes #3313 

---

## Type of Change

- [ ] 📝 Documentation update
- [ ] 🔒 Security fix
- [ ] 🧪 Tests only

---

## Checklist


Priyanshu-byte-coder:
This adds TypeScript syntax to a `.js` file, which is why **Playwright smoke tests** is failing:

```js
// e2e/dashboard-widgets.spec.js
const contributionRequests: string[] = [];
const goalPosts: unknown[] = [];
function mockMetricResponse(url: string) {
```

Playwright picks its esbuild loader from the file extension, so type annotations in a `.js` file are a parse error rather than something that gets stripped.

Two of the changes in here are genuinely good and worth keeping:

- Scoping the goals route to `/\/api\/goals(\?|$)/` so it stops shadowing `/api/goals/sync` — that's a real bug, the endpoint exists.
- `expect.poll(() => goalPosts.length).toBe(1)` instead of polling the array reference — better failure output.

To land them: either drop the type annotations, or rename the file to `.spec.ts` and keep them. Renaming is the nicer outcome long-term, but it's a bigger diff — your call.

One more thing: the diff removes the `await page.getByRole("button", { name: "Show 90-day range" }).first().click();` line and leaves the following `await page.locator(...)` chain, which changes what that test clicks. If that was intentional, say so in the description; if not, it's worth a second look.

PR #3316, opened 2026-07-31
Update api.spec.ts

### LINKED ISSUE #3315: [BUG] Fix authenticated API tests failing due to missing NextAuth HKDF salt
### Bug Description

NextAuth uses an HKDF (HMAC-based Extract-and-Expand Key Derivation Function) to derive the encryption key for JWT session tokens. When Next.js API routes decode the cookie on the server using getServerSession or getToken, they inherently use the session cookie name (e.g., "next-auth.session-token") as the salt parameter.

However, the standalone encode utility from next-auth/jwt defaults to an empty string ("") for the salt if one is not explicitly provided. Because buildSessionCookie() omitted the salt argument, the generated mock token was encrypted with a different key than the server expected. When Playwright sends an HTTP request (via page.context().request.get(...)) with this cookie, the Next.js API route fails to decrypt it, treats the user as unauthenticated, and incorrectly returns a 401 Unauthorized.

### Steps to Reproduce

Run the Playwright test suite targeting the API endpoints.

The "[API E2E] /api/metrics/contributions returns 200 with valid session cookie" test executes and generates a token via buildSessionCookie().

encode() derives the encryption key using the default empty salt "".

Playwright attaches the improperly salted cookie and directly invokes the Next.js backend API route.

The server's NextAuth logic attempts to decrypt the cookie using "next-auth.session-token" as the salt.

Decryption fails due to the mismatched HKDF salts, yielding an unauthenticated state (null session).

The API route rejects the request and returns a 401 Unauthorized status, causing the expect(status).not.toBe(401) assertion to fail.

### Affected Area

Other

### Screenshots

_No response_

### Browser & OS

_No response_

### Environment

None

### Additional Context

_No response_

rudrakshtank:
## Summary

Closes #3315

---

## Type of Change

- [ ] 📝 Documentation update
- [ ] 🔒 Security fix
- [ ] 🧪 Tests only

---

## Checklist

### DIFF
--- a/e2e/api.spec.ts
+++ b/e2e/api.spec.ts
@@ -26,6 +26,7 @@ async function buildSessionCookie(): Promise<string> {
       accessToken: "mock-access-token",
     },
     maxAge: 60 * 60,
+    salt: "next-auth.session-token",
   });
 }
 


Priyanshu-byte-coder:
What does the `salt` addition fix? As far as I can tell it's a no-op.

In this version of next-auth, `JWTEncodeParams.salt` is documented as: *"When no `salt` is passed, we assume this is a session token."* Since `"next-auth.session-token"` is exactly that default, passing it explicitly produces the same token as omitting it. The visual-regression suite encodes a session cookie the same way without `salt` and authenticates fine.

If it's fixing an actual failure you hit, I'd like to merge it — just add a line to the description saying what broke, because the diff doesn't show it. If it's defensive explicitness, that's a fair argument too, but say so.

The trailing-newline fix at the end of the file is a straightforward improvement either way.

PR #3318, opened 2026-07-31
Update check-deps.js

### LINKED ISSUE #3317: [BUG] Fix JSONC parsing for TS configs and wildcard package export resolution
### Bug Description

The dependency checker script contains two major flaws causing false positives for valid imports:

    Strict JSON Parsing for Configs: It uses standard JSON.parse to read tsconfig.json and jsconfig.json. Since TypeScript configuration files commonly contain comments and trailing commas (JSONC), JSON.parse crashes. This causes the script to silently fail internal alias resolution, subsequently flagging valid local alias imports (e.g., @/components/Button) as missing third-party dependencies.

    Incomplete Subpath Exports Resolution: The isValidPackageSubpath function does not correctly handle advanced wildcard exports (e.g., "./utils/*") or root conditional exports (e.g., "exports": { "import": "...", "require": "..." }). This causes valid deep imports from modern third-party modules to be incorrectly marked as invalid/missing.

### Steps to Reproduce

Create a tsconfig.json with standard comments (often auto-generated by Next.js) and custom path aliases.

Import a local file using a path alias (e.g., import Utils from "@/utils/helpers").

Run the script; observe it incorrectly flags @/utils/helpers as a missing dependency because it failed to load the tsconfig aliases.

Install a modern package with specific wildcard exports (like "exports": { "./features/*": "./dist/features/*.js" }) or root conditions.

Import from it (import { x } from 'pkg/features/auth').

Run the script; observe it incorrectly flags the valid subpath as missing.

### Affected Area

Other

### Screenshots

_No response_

### Browser & OS

_No response_

### Environment

None

### Additional Context

_No response_

rudrakshtank:
## Summary

Closes #3317 

---

## Type of Change

- [ ] 📝 Documentation update
- [ ] 🔒 Security fix
- [ ] 🧪 Tests only

---

## Checklist

### DIFF
--- a/scripts/check-deps.js
+++ b/scripts/check-deps.js
@@ -64,8 +64,18 @@ function extractImports(src) {
 
   return imports;
 }
+
+// Helper to parse JSON files with comments and trailing commas (JSONC) commonly found in tsconfig.json
+function parseJsonc(content) {
+  const clean = content
+    .replace(/\/\*[\s\S]*?\*\//g, "")
+    .replace(/(^|[^:])\/\/.*$/gm, "$1")
+    .replace(/,\s*([\]}])/g, "$1");
+  return JSON.parse(clean);
+}
+
 function loadConfigWithExtends(configPath) {
-  const config = JSON.parse(
+  const config = parseJsonc(
     fs.readFileSync(configPath, "utf8")
   );
 
@@ -98,6 +108,7 @@ function loadConfigWithExtends(configPath) {
     },
   };
 }
+
 function loadInternalAliases(rootDir) {
   const aliases = ["@/", "~/", "src/"];
 
@@ -128,6 +139,7 @@ function loadInternalAliases(rootDir) {
 
   return aliases;
 }
+
 function isValidPackageSubpath(pkgName, mod, cwd) {
   try {
     let pkgJson;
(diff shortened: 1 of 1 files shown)


Priyanshu-byte-coder:
I checked this against the actual repo and it fixes a problem that isn't happening, while introducing one that would.

**`check-deps.js` currently works.** On `main`:

```
$ node scripts/check-deps.js
✓  All imports accounted for (416 files checked)
```

`tsconfig.json` in this repo has no comments and no trailing commas, so `JSON.parse` handles it fine.

**`parseJsonc` breaks it.** The block-comment regex `/\/\*[\s\S]*?\*\//g` is not comment-aware, so it matches across the glob patterns in `paths` and `include`:

```
$ node -e "...parseJsonc(fs.readFileSync('tsconfig.json','utf8'))"
parseJsonc BREAKS tsconfig -> Expected ':' after property name in JSON at position 549 (line 30 column 14)
```

It starts eating at the `/*` inside `"@/*": ["./src/*"]` and runs on until it finds a `*/`, swallowing real config on the way. So this change turns a passing script into a failing one.

The line-comment regex has the same class of problem — `(^|[^:])\/\/` only avoids stripping after a colon, which saves `https://` by luck but would still corrupt a path like `"a//b"`.

If you do want JSONC support here (reasonable in principle, since `tsconfig.json` is allowed to have comments even though this one doesn't), it needs a real tokenizer that tracks string state — `strip-json-comments` or `jsonc-parser` both do it correctly in a few lines. A regex pass over JSON can't distinguish a comment from the same characters inside a string.

Leaving open in case you want to take that route.

PR #3272, opened 2026-07-29
fix: add HTTP status check before JSON parse

saurabhhhcodes:
This fix improves error handling to prevent silent failures.

PR #3322, opened 2026-07-31
fix: code quality and safety improvements

saurabhhhcodes:


Priyanshu-byte-coder:
Closing — this branch does not build. It contains changes I have already explained and closed twice (#3272/#3274 on 2026-07-29, #3279–#3300 on 2026-07-31):

- `parseInt(searchParams.get("days", 10) ?? "90", 10)` — `URLSearchParams.get()` takes **one** argument. This is a type error, and `parseInt` already had its radix in the correct position.
- `.sort((a, b) => a - b)` on arrays of **strings** — subtracting strings yields `NaN`, so the comparator is meaningless and sort order becomes unspecified.
- `merged_at === null` in place of `== null` — a regression: `== null` matches both `null` and `undefined`, and `merged_at` is optional here, so strict equality silently miscounts.

Type check and Build are red on this PR. Please run `pnpm type-check` locally before opening anything.

PR #2883, opened 2026-06-27
feat: Add SWR local storage cache layer for offline dashboard resiliency

dhruvil-1207:
Fixes #2801 

### Description
Implemented a lightweight client-side data resiliency layer using a Stale-While-Revalidate (SWR) model for the core dashboard widgets. 

This ensures that if a developer works on a spotty local network or encounters GitHub API rate-limits, the dashboard avoids crashing or showing endless loading spinners by instantly hydrating from a local snapshot while syncing live data in the background.

### Changes Made
- **PRMetrics, StreakTracker, and GoalTracker**: Intercepted incoming API dashboard metrics payloads.
- Added `localStorage` caching with timestamps.
- Implemented immediate UI hydration on load using the saved snapshot.
- Added silent background fetching to fetch live data and overwrite the stale snapshot without disrupting the user layout.
- Handled errors gracefully so the dashboard doesn't crash if `localStorage` is completely empty or disabled.

### Testing
- Tested locally by simulating an "Offline" network in Chrome DevTools.
- Verified that widgets instantly display cached data on hard refresh.
- Confirmed that failed background requests are quietly swallowed, preventing UI crashes.

PR #3348, opened 2026-08-02
feat: add SWR local storage cache layer for offline dashboard resiliency (#2801)

### LINKED ISSUE #2801: [FEAT] Add Stale-While-Revalidate (SWR) Local Storage Cache Layer for Offline Dashboard Resiliency
### Problem Statement

When developers work on a spotty local network, travel, or encounter temporary GitHub API rate-limiting blocks, the core dashboard widgets can freeze, display heavy loading spinners, or drop into empty un-hydrated states. There is currently no persistent client-side state backup to display immediate fallback metric snapshots.

### Proposed Solution

Implement a lightweight client-side data resiliency layer using a Stale-While-Revalidate (SWR) model:
1. **Local Snapshot Storage:** Intercept successful incoming API dashboard metrics payloads (commit streaks, PR tracking analytics, goals) and store them as a stringified JSON snapshot in `localStorage` with a timestamp.
2. **Instant Hydration & Silent Sync:** On initial dashboard load, immediately populate UI cards with the saved localStorage snapshot if available. Concurrently dispatch a background fetch request to sync live data and silently update the layout state once resolved.

### Feature Area

Dashboard

### Alternatives Considered

* **Forcing strict loading states:** Keeping the layout blocked behind a continuous loading spinner panel until a live network bridge resolves, which hurts UX speed and breaks accessibility during offline sessions.

### Acceptance Criteria

- [ ] Dashboard displays the last cached metrics instantly on hard refresh without a prolonged blank loading state.
- [ ] Toggling the browser network connection to "Offline" retains existing metric values instead of crashing widget components.
- [ ] Background network requests silently overwrite the stale local storage baseline metrics once internet connectivity re-establishes.
- [ ] Proper handling ensures zero runtime crashes if localStorage is completely empty or disabled by browser flags.

### Additional Context

This is an isolated client-side enhancement focused purely on developer experience (DX) and frontend speed optimization. I am a GSSoC '26 contributor and am ready to build this caching wrapper cleanly. Could you please review and assign this issue to me?

yachikadev:
## Summary

Implements a Stale-While-Revalidate (SWR) local storage cache layer for the dashboard's Streak Tracker and Goal Tracker widgets, so users see their last-known data instantly on load instead of a blank/loading state — even on spotty networks, during GitHub API rate-limiting, or fully offline.

Closes #2801

## Changes

- **`src/lib/localCache.ts`** (new) — lightweight `saveSnapshot` / `getSnapshot` / `clearSnapshot` helpers backed by `localStorage`, with `try/catch` guards so a full, disabled, or missing `localStorage` never throws at runtime.
- **`src/components/StreakTracker.tsx`** — on mount, hydrates instantly from the last cached streak/contribution snapshot before the network call resolves. On successful fetch, persists the new snapshot. On fetch failure, falls back to the cached snapshot instead of showing an error.
- **`src/components/GoalTracker.tsx`** — same pattern applied to the goals list: instant hydration from cache, snapshot saved on successful load, cached goals shown if the live fetch fails.

## How it works

1. **Instant hydration** — a `useEffect` on mount reads the cached snapshot from `localStorage` and populates state immediately, skipping the loading spinner if data is available.
2. **Background sync** — the existing fetch logic runs unchanged in parallel; on success it silently overwrites both the UI state and the cached snapshot.
3. **Graceful fallback** — if the live request fails (offline, rate-limited, etc.), the cached snapshot is shown instead of an error message, whenever one exists.

## Notes

- Scoped to the two widgets driven by `/api/metrics` and `/api/goals` (streaks + goals), matching the issue's acceptance criteria. Other widgets are untouched and can adopt the same `localCache.ts` utility in future PRs if desired.

### DIFF
--- a/src/components/GoalTracker.tsx
+++ b/src/components/GoalTracker.tsx
@@ -9,6 +9,7 @@ import { buildPublicGoalShareUrl } from "@/lib/goals/share";
 import GoalHistory from "@/components/GoalHistory";
 import EmptyState from "@/components/EmptyState";
 import WidgetSkeleton, { SkeletonBlock } from "./WidgetSkeleton";
+import { saveSnapshot, getSnapshot } from "@/lib/localCache";
 
 
 type Recurrence = "none" | "weekly" | "monthly";
@@ -90,9 +91,12 @@ export function useGoalTracker() {
     const data: { goals: Goal[] } = await response.json();
     const fetchedGoals = data.goals ?? [];
     setGoals(fetchedGoals);
+
+    // SWR: persist fresh snapshot for offline / instant-hydration use
+    saveSnapshot("goals", fetchedGoals);
+
     return fetchedGoals;
   }, []);
-
   const handleSync = useCallback(async () => {
     setSyncing(true);
     setSyncError(null);
@@ -131,6 +135,15 @@ export function useGoalTracker() {
     }
   }, [loadGoals]);
 
+ // SWR Step 1: instant hydration from localStorage before network resolves
+  useEffect(() => {
+    const cached = getSnapshot<Goal[]>("goals");
+    if (cached) {
+      setGoals(cached.data);
+      setLoading(false); // show stale goals immediately, no spinner
+    }
+  }, []);
+
   useEffect(() => {
     loadGoals()
       .then(async (fetchedGoals) => {
@@ -145,7 +158,14 @@ export function useGoalTracker() {
         }
       })
       .catch(() => {
-        setSyncError("Failed to load goals. Please try again.");
+        // SWR fallback: agar live fetch fail ho, cached snapshot dikhta rahe
+        const cached = getSnapshot<Goal[]>("goals");
+        if (cached) {
+          setGoals(cached.data);
+          setSyncError(null);
+        } else {
+          setSyncError("Failed to load goals. Please try again.");
+        }
       })
       .finally(() => {
         setLoading(false);
--- a/src/components/GoalTracker.tsx
+++ b/src/components/GoalTracker.tsx
@@ -158,7 +158,7 @@ export function useGoalTracker() {
         }
       })
       .catch(() => {
-        // SWR fallback: agar live fetch fail ho, cached snapshot dikhta rahe
+        // SWR fallback: if the live fetch fails, keep showing the cached snapshot
         const cached = getSnapshot<Goal[]>("goals");
         if (cached) {
           setGoals(cached.data);
--- a/src/components/StreakTracker.tsx
+++ b/src/components/StreakTracker.tsx
@@ -120,7 +120,7 @@ export function useStreakTracker() {
     } catch (err) {
       console.error("Failed to fetch streak data:", err);
 
-      // SWR fallback: agar live fetch fail ho, cached snapshot dikhado
+     // SWR fallback: if the live fetch fails, fall back to the cached snapshot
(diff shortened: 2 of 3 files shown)


Priyanshu-byte-coder:
Good PR — the stale-while-revalidate approach is right: instant hydration from `localCache`, fresh snapshot persisted after each fetch, and falling back to the cached snapshot instead of showing an error when the network fails. CI is green and `src/lib/localCache.ts` is a clean, reusable helper. (Note #2883 attempted the same thing and is now closed, so this is the one I want.)

One thing to fix before I merge — two inline comments are in Hindi:

```
// SWR fallback: agar live fetch fail ho, cached snapshot dikhta rahe   (GoalTracker.tsx)
// SWR fallback: agar live fetch fail ho, cached snapshot dikhado       (StreakTracker.tsx)
```

Code comments in this repo are English-only so every contributor can read them. Please reword those two (e.g. `// SWR fallback: if the live fetch fails, keep showing the cached snapshot`) and I'll merge it.

yachikadev:
Hi @Priyanshu-byte-coder , just following up on this PR — it's been open for a while now. Could you review it when you have a moment? Happy to make any changes you'd suggest, or if it's good to go, would appreciate the merge. Thanks!

PriyanshuValura:
Nice shape overall — `localCache.ts` is tidy (SSR guard, try/catch, typed, prefixed keys), and the SWR sequence is right: hydrate from cache, fetch, persist. Keying the streak snapshot per account (`streak-${selectedAccount}`) is a good catch that's easy to miss.

Three things stop me merging it as-is. Two are correctness, not polish.

**1. The snapshot never expires.** `CachedSnapshot` stores `timestamp`, but nothing ever reads it. So a snapshot written three months ago hydrates on mount and calls `setLoading(false)` — the user sees a stale streak with no spinner and no indication it's old. It gets worse in the failure path: on a fetch error you `setError(null)` and keep showing the cached copy, so months-old numbers render as if they were live. A dashboard quietly showing wrong figures is worse than one showing an error.

You already have the field to fix it:

```ts
export function getSnapshot<T>(key: string, maxAgeMs = 24 * 60 * 60 * 1000) {
  // ...
  if (Date.now() - parsed.timestamp > maxAgeMs) return null;
  return parsed;
}
```

And on the error path, surfacing something like "showing data from 2 hours ago" would be better than silently swallowing the error.

**2. Nothing clears the cache on sign-out.** `clearSnapshot` is exported but never called anywhere in the diff. `localStorage` is per-origin, not per-user, so if two people use DevTrack in the same browser — a shared laptop, or someone switching GitHub accounts — the second user's first paint hydrates the first user's goals and streak before the network responds. `SignOutButton` calls `signOut({ callbackUrl: "/" })` and touches nothing else.

Either clear the `devtrack_cache_` keys on sign-out, or fold the user id into the key the way you already did for account. Keying by user is probably the sturdier of the two.

**3. Needs a rebase.** The `Continuous Integration Security & Audit Suite` failure is stale — that job was red repo-wide when you opened this and is green on `main` now. A rebase should clear it.

Minor, non-blocking: `saveSnapshot` has no `typeof window` guard while `getSnapshot` does. The try/catch covers it in practice, but the asymmetry looks unintentional. Also worth adding a small test for `localCache` — expiry and the malformed-JSON path are exactly the cases that break quietly later.

Fix 1 and 2 and I'll merge this.

PR #3505, opened 2026-08-25
fix: Fix landing GitHub stats showing zeros

### LINKED ISSUE #3493: [BUG] Live GitHub stats show 0 instead of real values
### Bug Description

The landing page advertises "Live GitHub Signals" pulled from the repo, but several stat widgets render 0 instead of real numbers: COMMITS IN REPO, PRS MERGED, CONTRIBUTORS, GITHUB STARS, FORKS, and GOOD FIRST ISSUES.

This contradicts the surrounding copy, which states "19 issues are tagged good first issue and waiting right now" directly next to a GOOD FIRST ISSUES stat showing 0. The repo actually has 44 stars, real forks, and 20+ contributors shown in the avatar grid, so the live numbers clearly aren't being fetched (or the fetch fails and silently falls back to 0) on the deployed site.

Expected: stats should reflect live values from the GitHub API, matching the repo's actual stars/forks/contributors/good-first-issue counts.

### Steps to Reproduce

1. Visit https://devtrack-silk-kappa.vercel.app/
2. Scroll to the stats row below the contributions heatmap
3. Scroll to the "OPEN SOURCE / BUILT IN PUBLIC" section near the contributor avatar grid
4. Observe all stat values read "0" or "0+", while the adjacent text says "19 issues are tagged good first issue"

### Affected Area

Landing Page

### Screenshots

_No response_

### Browser & OS

_No response_

### Environment

None

### Additional Context

_No response_

VedantMadane:
## Summary

Count open good-first-issues using GitHub Search API with `is:issue` to exclude pull requests and read `total_count`, and include a standard `User-Agent` header on GitHub API requests.

## Changes

- Count good-first-issues via search query (`repo:Priyanshu-byte-coder/devtrack label:"good first issue" is:issue is:open`) reading `total_count` rather than listing up to 100 issues where pull requests are mixed in
- Send explicit `User-Agent` header (`devtrack-landing-stats`) on GitHub API calls as standard client hygiene

Fixes #3493


Priyanshu-byte-coder:
I dug into this one properly because the diagnosis is testable, and it splits into one change I want and one built on a premise that doesn't hold.

**The User-Agent part doesn't fix anything.** The claim is that unauthenticated fetches 403 without a `User-Agent`. GitHub does require one — but Node's `fetch` (undici, which is what Next uses server-side) already sends a default, so the header is never actually absent:

```
$ node -e "fetch('https://api.github.com/repos/Priyanshu-byte-coder/devtrack', {headers:{Accept:'application/vnd.github.v3+json'}}).then(r=>console.log(r.status))"
200
```

I can reproduce a 403 only by explicitly blanking the header (`curl -H "User-Agent:"`), which isn't what the app does. Setting it is harmless and arguably good hygiene, but it isn't the cause of the zeros.

**The good-first-issue change is a real improvement — keep it.** `/issues?labels=...` returns pull requests alongside issues (GitHub treats PRs as issues on that endpoint), and `per_page=100` silently caps the count. Moving to `search/issues` with `is:issue` and reading `total_count` fixes both. No PRs currently carry the label so the number is the same today (19 either way, I checked), but it's correct rather than accidentally correct.

**What I think is actually happening.** Running the page's exact four-call sequence unauthenticated:

```
repo         200  x-ratelimit-remaining=47
contributors 200  x-ratelimit-remaining=48
gfi search   200  x-ratelimit-remaining=8
prs search   200  x-ratelimit-remaining=9
```

Core API is 60/hr; **Search is 10/min** unauthenticated, and this page makes two search calls per render. That's the quota that runs out first, and when either search 403s the values fall back to zero. Note this PR adds a second search call where there was one, so it makes that pressure slightly worse, not better.

The fix for #3493 is almost certainly setting `GITHUB_TOKEN` in the deployment environment — it's already read at the top of `fetchRepoStats` and lifts search to 30/min and core to 5,000/hr. Worth confirming whether it's set in production before changing more code.

**What I'd like:** keep the good-first-issue counting change, keep the `User-Agent` if you want it, but please update the PR description so it doesn't claim the UA fixes the zeros — the next person debugging this will follow that trail. And if you can check whether `GITHUB_TOKEN` is configured on the deployment, that's the thing most likely to actually close #3493.

Verified against current `main`: merges clean, type-check clean, 163 files / 2387 tests passing, build green.

